PR #1, opened 2020-09-09
Fixed a compatibility issue with WP 5.5 and newer versions of GFUR.

eihabi:
This PR adds the [custom activation template](https://gravitywiz.com/customizing-gravity-forms-user-registration-activation-page/) snippet to the repo and fixes an issue with newer versions of GFUR where the `page` parameter has been changed to `gfur_activation`.

Ticket: [#19745](https://secure.helpscout.net/conversation/1275451153/19745?folderId=3808239)

### DIFF
--- a/gravity-forms/gw-custom-activation-template.php
+++ b/gravity-forms/gw-custom-activation-template.php
@@ -3,16 +3,17 @@
 * Gravity Forms Custom Activation Template
 * http://gravitywiz.com/customizing-gravity-forms-user-registration-activation-page
 */
-add_action('wp', 'custom_maybe_activate_user', 9);
+add_action( 'wp', 'custom_maybe_activate_user', 9 );
 function custom_maybe_activate_user() {
 
-    $template_path = STYLESHEETPATH . '/gfur-activate-template/activate.php';
-    $is_activate_page = isset( $_GET['page'] ) && $_GET['page'] == 'gf_activation';
-    
-    if( ! file_exists( $template_path ) || ! $is_activate_page  )
-        return;
-    
-    require_once( $template_path );
-    
-    exit();
+	$template_path    = STYLESHEETPATH . '/gfur-activate-template/activate.php';
+	$is_activate_page = isset( $_GET['page'] ) && $_GET['page'] === 'gf_activation';
+
+	if ( ! file_exists( $template_path ) || ! $is_activate_page ) {
+		return;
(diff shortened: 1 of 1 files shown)


gravity-forms/gw-custom-activation-template.php
```diff
@@ -0,0 +1,20 @@
...
+/**
+* Gravity Forms Custom Activation Template
+* http://gravitywiz.com/customizing-gravity-forms-user-registration-activation-page
+*/
+add_action( 'wp', 'custom_maybe_activate_user', 9 );
+function custom_maybe_activate_user() {
+
+	$template_path    = STYLESHEETPATH . '/gfur-activate-template/activate.php';
+	$is_activate_page = isset( $_GET['page'] ) && $_GET['page'] === 'gf_activation';
+	$is_activate_page = $is_activate_page || isset( $_GET['gfur_activation'] ); // WP 5.5 Compatibility
```
spivurno, line 11: There is a potential for a notice here if the the page parameter is not set on line 11. Thoughts on combining these two lines like so?

```
$is_activate_page = isset( $_GET['page'] ) && in_array( $_GET['page'], array( 'gf_activation', 'gfur_activation' ), true );
```

spivurno (changes requested):

spivurno:
@eihabi Just a note that we will need to make this change in a few of our perks as well. I have a card for it here and it's assigned to me. If we get a ticket for it before I'm able to tackle it, feel free to update the card with whichever perks you fix it for in the given ticket.

https://www.notion.so/gravitywiz/GF-User-Registration-Updates-bb203c4643bd46a1a4b1220b0ce363ae

gravity-forms/gw-custom-activation-template.php
eihabi: @spivurno I thought the change was doing away with the `page` parameter in favor of a new `gfur_activation` parameter, is this incorrect?

If page parameter is empty or not set to `gf_activation`, line 11 would just check for the existence of the `gfur_activation` parameter and set `is_activate_page` to `true`, in all other cases it should be `false` and the custom template wouldn't be loaded.

spivurno: Dude, you're 💯 % right. Please ignore me. 😆

spivurno (approved):

PR #2, opened 2020-09-10
Added a new snippet to disable all form inputs with the submit button.

eihabi:
This is a new snippet to disable all fields on a form when GPPA is querying the back-end and the submit button is disabled.

It uses `gppa_disable_form_navigation_toggling` to detect when GPPA is about to disable the submit button and immediately disables any form element that is not currently disabled (while also tagging such elements with a `data-gppa-disabled` attribute).

Once GPPA is done and it is about to re-enable the submit button, the same filter is checked and this snippet reverses its actions.


claygriffiths (approved):
Good stuff! I initially was going to leave a comment about the fact that it disables fields for all forms on the page, but that appears to be the goal based on on the PR title and code comments 😃

PR #3, opened 2020-09-14
Updated `gw-gravity-forms-cache-buster.php` to 0.3 which passes URL params.

eihabi:
@spivurno Just heads up, this was already reviewed but I think it got lost in the snippet-library creation fog. I'm merging it right away and sending to Scott for an issue.

### DIFF
--- a/gravity-forms/gw-cache-buster.php
+++ b/gravity-forms/gw-cache-buster.php
@@ -4,7 +4,7 @@
  *
  * Bypass your website cache when loading a Gravity Forms form.
  *
- * @version 0.2
+ * @version 0.3
  * @author  David Smith <david@gravitywiz.com>
  * @license GPL-2.0+
  * @link    http://gravitywiz.com/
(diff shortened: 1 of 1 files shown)


spivurno:
Perfect. Thanks, @eihabi!

ouuan:
I suspect that this is vulnerable to XSS attacks since the GET params are directly used in the output. However, I have no license for either Gravity Perks or Gravity Forms so I did not test whether it is actually exploitable yet.

I have already reported this to @spivurno via email but received no response yet, so I am posting here, hoping to get a response. If this is actually vulnerable, I am willing to write a patch to fix it and help create a [security advisory](https://docs.github.com/en/code-security/security-advisories/repository-security-advisories).

spivurno:
Hey @ouuan, thanks for reaching out! 

I've reviewed the code but there doesn't appear to be any security issue here. The POST request that the GET parameters are passed through to is a public endpoint and as such anyone can post any parameters. We do not control that endpoint, only utilize it to fetch the form markup. The endpoint itself is responsible for sanitizing the data it receives.

ouuan:
Hi! I think you misunderstood me, but I don't know whether it's OK to discuss the vulnerability details here in public. Can we discuss via email instead?

spivurno:
@ouuan Sure! I received your follow up email and will discuss with the team. Thank you. 🙂

PR #4, opened 2020-09-23
Rounding by Increments Update

eihabi:
This PR imports the Rounding by Increments snippet and implements the following updates:

- Prevent rounding of floats if entered number is not greater than max
- Prevent rounding of floats if entered number is not less than min
- Adds support for rounding by decimal value
- Adds `gw-rounding-custom` filter to allow end-users to implement rounding algorithm of choice
- General PHPCS house-keeping

Additional info in HS.

### DIFF
--- a/gravity-forms/gw-gravity-forms-rounding.php
+++ b/gravity-forms/gw-gravity-forms-rounding.php
@@ -28,7 +28,7 @@ class GW_Rounding {
 
 	protected static $is_script_output = false;
 
-	protected $class_regex = 'gw-round-(\w+)-?(\w+)?';
+	protected $class_regex = 'gw-round-([\w|0-9.]+)-?([\w|0-9.]+)?';
 
 	public static function get_instance() {
 		if ( null == self::$instance ) {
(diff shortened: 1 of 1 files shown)


spivurno:
@eihabi Awesome work!

I love the idea for the custom filter. I think it'd be awesome to make it more generic/flexible. Rather than checking specifically for the "custom" string, we could pass any non-standard "modifier" through with a dynamic filter name.

```js
value = window.gform.applyFilters( 'gw_rounding_{0}'.format( action ), value, actionValue );
```

So the user might have a custom class like `gw-rounding-nearest9` and then bind to the dynamic `gw_rounding_nearest9` filter. This would allow the user to assign more meaning to their custom rounding class names and support multiple custom classes names without needing to know which field is being processed.

Thoughts?

spivurno (changes requested):
Forgot to leave my previous comment as a review. Let me know what you think about the idea of a open-ended custom class powered by a dynamic filter.

eihabi:
@spivurno Thanks and great idea! I've updated the filter to allow for arbitrary names.

spivurno (changes requested):
A clever and clean implementation. Good work! 

One last thing... for security, we'll need to support this filter on the PHP side of things as well. See the `GW_Rounding::round()` method.

eihabi:
@spivurno Done! Thanks for catching that, I overlooked the back-end on this one 😅

spivurno (approved):
Awesome work, my friend. 🌮

PR #5, opened 2020-09-29
Fixed filter name to match the standard rounding class names.

eihabi:
This PR renames the filters from `gw_rounding_xxx` to `gw_round_xxx` to match the rest of the filters and class names.

It also addresses an issue where a custom filter without an actionValue set in the class name would default to the `round` method (e.g. `gw-round-nearest9` would be broken while `gw-round-nearest-9` would work).


spivurno (approved):
Looks good!

PR #6, opened 2020-10-12
Added a parameter to pass a title for coupon code

sbassah:
An update for the snippet to pass a field id to get a value that will be used to set the coupon name.

### DIFF
--- a/gravity-forms/gw-create-coupon.php
+++ b/gravity-forms/gw-create-coupon.php
@@ -48,7 +48,14 @@ class GW_Create_Coupon {
 		}
 
 		$coupon_code = rgar( $entry, $this->_args['source_field_id'] );
-		$coupon_name = rgar ($entry, $this->_args['name_field_id']);
+
+		if ($this->_args['name_field_id'] == false){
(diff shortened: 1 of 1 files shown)


spivurno (changes requested):
1. Does this work for WooCommerce as well?
2. What happens if field is specified or the field is empty? I would expect that it should default back to the coupon code.

sbassah (commented):
Per your review, I've updated the snippet to first check if an ID was passed for the name field, if not the coupon code is used as the coupon name. Also, if a name field ID is passed, this update checks, if the name field is blank, and if it's blank it uses the coupon code as the coupon name.

With WooCommerce, the Post Title is the coupon code, so I assigned the coupon name to the Post content.

spivurno (approved):
Looks good! 

Approved but could you make one change before you merge? I think an inline comment explaining why we pass the coupon code as the post title and the coupon name is passed as the post content in the WC function would be helpful for understanding this in the future.

spivurno:
Also, good work! 🤘

PR #7, opened 2020-10-16
Support to set the coupon limit using a field value

sbassah:
PR to add support to set the coupon limit using a field value. If the value entered for the coupon_limit parameter is not a valid field ID, the parameter value will be used as the coupon limit.

### DIFF
--- a/gravity-forms/gw-create-coupon.php
+++ b/gravity-forms/gw-create-coupon.php
@@ -56,10 +56,10 @@ class GW_Create_Coupon {
 			$coupon_name = ($coupon_name == ''? $coupon_code : $coupon_name);
 		}
 
-		$limit_value = rgar($entry, $this->_args['meta']['coupon_limit']);
-		if ($limit_value != null){
-			 $this->_args['meta']['coupon_limit'] = $limit_value;
-		}
+		$limit_value = $this->_args['meta']['coupon_limit'];
+        if( is_callable( $limit_value ) ) {
+            $limit_value = call_user_func( $limit_value );
+        }
 
 		$amount      = $this->_args['amount'];
 		$type        = $this->_args['type'];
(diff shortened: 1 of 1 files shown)


gravity-forms/gw-create-coupon.php
```diff
@@ -56,6 +56,11 @@ public function create_coupon( $entry, $form ) {
 			$coupon_name = ($coupon_name == ''? $coupon_code : $coupon_name);
 		}
 
+		$limit_value = rgar($entry, $this->_args['meta']['coupon_limit']);
```
spivurno, line 59: If the user wants to actually set a static limit there is a chance that it will match a field ID and be incorrectly replace. My instinct here is to allow this property to be a [callable function](https://www.php.net/manual/en/function.call-user-func.php) to which we can pass the $entry.

If you're not sure what I mean, let's implement it together! 🙂

spivurno (changes requested):

sbassah (commented):
Updated the snippet to use a callable function in setting the coupon limit. This way, a static limit which is similar to a field ID on the form can be used without any issue.

spivurno:
@sbassah So, forgive me, I made a few formatting changes to this snippet since you forked it so now there are merge conflicts. Not your fault and not hard to resolve but a little tricky to understand. 

I think the easiest path forward now is to close this PR and open a new PR from a new branch directly on this repo. Let me know if you need any help with that. 

Two other things.

1. The tabbing is a little off. I'd recommend installing our code sniffer which handles checking for things like this and automatically giving you a heads up. 

    Here's our internal documentation for getting this setup. I can help if you run into any issues.

    https://www.notion.so/gravitywiz/Configuring-PHP_CodeSniffer-for-all-Perks-b2bf33a9ce2041239a0bba26b473ffd2

2. Can you reach out to Scott and let him know about this change? You'll want to ask if he'd like you to implement updates for articles that are attached to snippets like this one or if he wants to handle it himself.

sbassah:
Closing PR to open a new one

PR #9, opened 2020-10-20
Add GPUID Exclude Chars Snippet

eihabi:
This PR adds the GPUID Exclude Chars snippet to the library. It also [fixes an issue](https://github.com/gravitywiz/snippet-library/commit/d3b7cdc8bdd87b04b9944c8896c101bf3f856cb7) with generated GPUIDs.

`array_diff()` doesn't reset array keys, which caused the random replace loop to fail sometimes and end up replacing an excluded character with a `null` causing the UID to be shorter than the specified length.

The fix uses `array_values()` to copy the generated array, ensuring that keys will be sequential.

### DIFF
--- a/gravity-forms/gw-unique-id-exclude-characters.php
+++ b/gravity-forms/gw-unique-id-exclude-characters.php
@@ -1,0 +1,84 @@
+<?php
+/**
+* Gravity Wiz // Exclude characters from GPUID
+*
+* Allows you to exclude characters from unique IDs generated using GPUID.
+*
+* @version   1.1
+* @author    Eihab Ibrahim <eihab@gravitywiz.com>
+* @license   GPL-2.0+
+* @link      http://gravitywiz.com/calculate-number-of-days-between-two-dates/
+* @copyright 2020 Gravity Wiz
+*/
+class GWExcludeUIDChars {
+	/**
+	 * GWExcludeUIDChars Constructor
+	 *
+	 * Expects an $args named array with the following keys=>values:
+	 * 'form_id':  ID of the GF form to perform replacements in
+	 * 'field_id': GPUID Input ID
+	 * 'exclude':  String containing characters to exclude
+	 * 'sitewide': [optional] Boolean, set to true to apply to all GPUIDs
+	 *
+	 * Example usage in functions.php:
+	 * new GWExcludeUIDChars( array(
+	 *     'form_id'  => 11,
+	 *     'field_id' => 1,
+	 *     'exclude'  => 'li0o'
+	 * ) );
+	 */
+	public function __construct( $args ) {
+		extract( wp_parse_args( $args, array(
+			'form_id'  => false,
+			'field_id' => false,
+			'exclude'  => false,
+			'sitewide' => false,
+		) ) );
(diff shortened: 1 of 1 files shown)


gravity-forms/gw-unique-id-exclude-characters.php
```diff
@@ -0,0 +1,83 @@
...
+	 *
+	 * Example usage in functions.php:
+	 * new GWExcludeUIDChars( array(
+	 *     'form_id'  => 11,
+	 *     'field_id' => 1,
+	 *     'exclude'  => 'li0o'
+	 * ) );
+	 */
+	public function __construct( $args ) {
+		extract( wp_parse_args( $args, array(
```
claygriffiths, line 31: One handy note regarding [`extract()`](https://www.php.net/manual/en/function.extract.php), if you use a PHP DocBlock with `@var`'s, PhpStorm will provide autocomplete and all of its goodness.

gravity-forms/gw-unique-id-exclude-characters.php
```diff
@@ -0,0 +1,83 @@
...
+			'form_id'  => false,
+			'field_id' => false,
+			'exclude'  => false,
+			'sitewide' => false,
+		) ) );
+		if ( ! $form_id || ! $field_id || ! $exclude ) {
+			if ( ! $sitewide || ! $exclude ) {
```
claygriffiths, line 38: Can this nested conditional be DRYed up any? `$exclude` appears to be repeated.

gravity-forms/gw-unique-id-exclude-characters.php
```diff
@@ -0,0 +1,83 @@
...
+		$this->form_id  = $form_id;
+		$this->field_id = $field_id;
+		$this->sitewide = $sitewide;
+		$this->exclude  = str_split( $exclude );
+
+		add_filter( 'gpui_unique_id', array( &$this, 'exclude_action' ), 10, 3 );
+	}
+
+	public function exclude_action( $unique, $form_id, $field_id ) {
+		if ( ! $this->sitewide && ( intval( $form_id ) !== $this->form_id || intval( $field_id ) !== $this->field_id ) ) {
```
claygriffiths, line 51: Methods like `should_run_action` (or similar) can be helpful in conditionals like these.

claygriffiths (approved):
Overall, LGTM! Just a few minor notes from me.

gravity-forms/gw-unique-id-exclude-characters.php
eihabi: Handy! I'll add it in future snippets/updates 👍

gravity-forms/gw-unique-id-exclude-characters.php
eihabi: The intention here was to allow for site-wide application or a form/field specific targeting. The second `$exclude` check is required just in case the first conditional fails on missing `$form_id` or `$field_id`.

gravity-forms/gw-unique-id-exclude-characters.php
eihabi: Noted! 👍

PR #13, opened 2020-11-20
Added support to set WooCommerce Coupons start date

sbassah:
This is a PR adding support to set a start date for Woocommerce Coupons. WooCommerce Coupons don't have a field or attribute to set the start date, so the start date value passed is assigned to the post_date variable. This means that, if the date is in the future it will schedule the coupon post but if it's in the past or it's the current date it will publish the coupon.

The start_date parameter is expected to be an element within the meta array and in this format; '2020-12-11'. If no start_date value is passed, the current date is used.

### DIFF
--- a/gravity-forms/gw-create-coupon.php
+++ b/gravity-forms/gw-create-coupon.php
@@ -185,7 +185,7 @@ class GW_Create_Coupon {
             $start_date = date('Y-m-d H:i:s');
         }
 
-        // WooCommerce coupon uses the Post Title as the coupon code hence $coupon_code is assiged to Post Title and $coupon_name is assigned to the Post Content
+        // WooCommerce coupon uses the Post Title as the coupon code hence $coupon_code is assigned to Post Title and $coupon_name is assigned to the Post Content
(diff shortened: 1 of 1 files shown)


spivurno:
@sbassah Great start!

1. Did you pull the latest before making your changes? I'm seeing a merge conflict and I'm wondering if you started with an outdated version of the snippet?
2. That tabbing looks little wonky. Do you have your PhpStorm code style setup to use WordPress? Here is [how mine is configured](https://gwiz.io/334SH1h).
3. ```$start_date = $this->_args['meta']['start_date'];``` 

    This will generate a notice if the `start_date` key is not set. I'd recommend using rgar() here which is a Gravity Forms helper function for checking for a value in an array that might not exist.

    ```$start_date = rgar( $this->_args['meta'], 'start_date' );```

sbassah:
I remember using git pull to get the latest version, but it looks like I didn't do it well. Sorry about that. I'm closing this PR and I will open a new one with the changes requested. Thanks

PR #14, opened 2020-11-21
Added Support to Set WooCommerce Coupon Start Date

sbassah:
**Initial Comment from first PR**
This is a PR adding support to set a start date for Woocommerce Coupons. WooCommerce Coupons don't have a field or attribute to set the start date, so the start date value passed is assigned to the post_date variable. This means that, if the date is in the future it will schedule the coupon post but if it's in the past or it's the current date it will publish the coupon.

The start_date parameter is expected to be an element within the meta array and in this format; '2020-12-11'. If no start_date value is passed, the current date is used.

Update
@spivurno I've made changes per your feedback in the first PR. Thanks


spivurno (approved):
Approved!

sbassah:
Thanks, David

PR #15, opened 2020-11-25
Added plugin header to gpld-exclude-blocked-dates-from-modifiers.php

scyt:
This PR adds a plugin header for gpld-exclude-blocked-dates-from-modifiers.php


spivurno (approved):
Approved with one change. Can you update the Plugin URI to the GP Limit Dates documentation?

PR #16, opened 2020-12-01
Converts gw-form-list-active-by-default.php to a plugin

scyt:
This PR converts the Form List Active by Default snippet to a plugin since it doesn't require any configuration.


spivurno (approved):
Looks good!

PR #17, opened 2020-12-02
Require List Columns

eihabi:
This PR imports the [Require All Columns of List Field](https://gravitywiz.com/require-all-columns-of-list-field/) snippet.

It also fixes an issue where validation would fail if the value passed is `0`. [`empty()`](https://www.php.net/manual/en/function.empty.php) returns true for `'0'` and `0` which is not the desired effect here.

HS: 21317


spivurno (approved):
Looks good!

PR #19, opened 2020-12-08
Added new snippet: gw-update-posts.php

scyt:
This PR adds a new snippet that will update WP post titles and content with values from GF entry. Usage example:

```
new GW_Update_Posts( array(
	'form_id' => 123,
	'post_id' => 1,
	'title'   => 2,
	'content' => 3
) );
```

### DIFF
--- a/gravity-forms/gw-update-posts.php
+++ b/gravity-forms/gw-update-posts.php
@@ -1,0 +1,60 @@
+<?php
+/**
+ * Gravity Wiz // Gravity Forms // Update Posts
+ *
+ * Update existing post title and content with values from Gravity Forms.
+ *
+ * @version 0.1
+ * @author  Scott Buchmann <scott@gravitywiz.com>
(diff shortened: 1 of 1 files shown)


spivurno (approved):
Looks good!

spivurno (changes requested):
Oops! Clay made a good point on security. We'll want to make sure the user has permission to edit the post before we update it. Otherwise, folks could override the field value with any post ID and update the content. 

**Edit:**
This is a good place to start.

```php
current_user_can( 'edit_post', $post->ID );
```

scyt:
I added a simple bail if they don't pass a cap check after the post is retrieved. Do you think that's sufficient, or did you have something more in mind?

spivurno:
That's perfect! 

Only other thing is a coding standard nitpick; comments should be sentenced cased.

✅

spivurno (approved):

PR #24, opened 2021-01-05
Assigned the ACF Repeater field name to a Variable.

sbassah:
Assigned the ACF Repeater field name to a variable to make it easy for users to configure the snippet. We missed this when we were tweaking the code.


spivurno (approved):
Good catch!

PR #25, opened 2021-01-06
Gpls disable rule for user role

sbassah:
Snippet to disable Limit Submission Rule for Specific User Role

### DIFF
--- a/gp-limit-submissions/gpls-disable-rule-for-user-role.php
+++ b/gp-limit-submissions/gpls-disable-rule-for-user-role.php
@@ -1,0 +1,14 @@
+<?php
+/**
+ * Gravity Perks // GP Limit Submissions // Disable Rule for a Specific User Role
+ * http://gravitywiz.com/documentaiton/gravity-forms-limit-submissions/
+ */
+// Change '123' to the ID of the Form
+add_filter( 'gpls_rule_groups_123', function( $rule_groups, $form_id ) {
+	$user = wp_get_current_user();
+	//Change administrator to the role that will use the form without limit
+	if ( in_array( 'administrator', (array) $user->roles ) ) {
+		$rule_groups = array();
+	}
+	return $rule_groups;
+}, 10, 2 );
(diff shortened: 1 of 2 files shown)


gp-limit-submissions/gpls-disable-rule-for-user-role.php
```diff
@@ -0,0 +1,14 @@
+<?php
+/**
+ * Gravity Perks // GP Limit Submissions // Disable Rule for a Specific User Role
+ * http://gravitywiz.com/documentaiton/gravity-forms-limit-submissions/
+ */
+// Change '123' to the ID of the Form.
+add_filter( 'gpls_rule_groups_123', function( $rule_groups, $form_id ) {
+	$user          = wp_get_current_user();
+	$role_to_check = 'administrator'; //Change to the slug of the role, i.e. administrator, editor, author, contributor, subscriber.
```
spivurno, line 9: Let's move this to the first line. Then, update the comment to be above the variable and change the word "Change" to "Update" to follow the same style we use with other snippets.

gp-limit-submissions/gpls-disable-rule-for-user-role.php
```diff
@@ -0,0 +1,14 @@
+<?php
+/**
+ * Gravity Perks // GP Limit Submissions // Disable Rule for a Specific User Role
+ * http://gravitywiz.com/documentaiton/gravity-forms-limit-submissions/
+ */
+// Change '123' to the ID of the Form.
```
spivurno, line 6: Let's use "Update" instead of "Change" here as well.

gp-limit-submissions/gpls-disable-rule-for-user-role.php
```diff
@@ -0,0 +1,14 @@
+<?php
+/**
+ * Gravity Perks // GP Limit Submissions // Disable Rule for a Specific User Role
```
spivurno, line 3: Nitpicky but since this will essentially dislabe all limit feeds for a given form, it might be a clearer to change the name to "Disable Limits for a Specific User Role". If you agree, let's update the file name as well.

spivurno (changes requested):
Good stuff! A few tweaks for usability and clarity. 🙂

sbassah:
I have submitted a PR for the changes requested for your review.

spivurno (approved):
Excellent work, my friend!

PR #27, opened 2021-02-09
Fix Custom Database Snippet Filtering

eihabi:
This PR overrides the `\GPPA_Object_Type_Database->constructor()` and 
`\GPPA_Object_Type_Database->add_filter_hooks()` to ensure that filtering hooks are set and fired correctly.

The default filter names are `gppa_pre_object_type_query_database` and `gppa_object_type_database_filter`, this breaks down when using this snippet as the custom class ID (`database-testing`) causes GPPA to attempt to apply `gppa_object_type_database-testing_filter` which in turn results in `\GPPA_Object_Type_Database->build_where_clause()` to never run.

Reference: #22473


eihabi:
Closing in favor of moving this to core (GPPA:127).

PR #29, opened 2021-02-23
Add GPPA Populate Child Entries

eihabi:
This PR migrates the [GPPA Populate Child Entries](https://gist.github.com/spivurno/dc425112b60c8aea98b7b947a50a8cf3) snippet to the library.

It also fixes an issue where GPNF's view model may not be ready on load due to the latest performance enhancements and lazy loading.

#22273

### DIFF
--- a/gp-populate-anything/gppa-populate-child-entries.php
+++ b/gp-populate-anything/gppa-populate-child-entries.php
@@ -126,20 +126,9 @@ class GPPA_Populate_Child_Entries {
 						return null;
 					}
 
-					// Wait for the first nested form field view model to be available
-					var waitTimeoutCount = 0;
-					self.initWait = function() {
-						var form = window[ 'GPNestedForms_{0}_{1}'.format( self.formId, self.nestedFormFieldIds[ 0 ] ) ];
-						if ( form && form.viewModel ) {
-							return self.init();
-						}
-						if ( waitTimeoutCount < 50 ) {
-							waitTimeoutCount++;
-							window.setTimeout(self.initWait, 1);
-						}
-					}
-
-					self.initWait();
+					$(function(){
+						self.init();
+					});
 
 				}
 
--- a/gp-populate-anything/gppa-populate-child-entries.php
+++ b/gp-populate-anything/gppa-populate-child-entries.php
@@ -21,7 +21,7 @@
  * 3. Select the desired child form for which you wish to fetch child entries.
  * 4. Add a filter and filter where Parent Entry ID entry meta is equal to the the Hidden field you created in Step 1.
  *
- * @version  1.1
+ * @version  1.2
  * @author   David Smith <david@gravitywiz.com>
  * @license  GPL-2.0+
  * @link     https://gravitywiz.com/documentation/gravity-forms-populate-anything/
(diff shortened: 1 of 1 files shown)


gp-populate-anything/gppa-populate-child-entries.php
```diff
@@ -126,7 +126,20 @@ public function output_script() {
...
+					// Wait for the first nested form field view model to be available
+					var waitTimeoutCount = 0;
+					self.initWait = function() {
+						var form = window[ 'GPNestedForms_{0}_{1}'.format( self.formId, self.nestedFormFieldIds[ 0 ] ) ];
+						if ( form && form.viewModel ) {
+							return self.init();
+						}
+						if ( waitTimeoutCount < 50 ) {
+							waitTimeoutCount++;
+							window.setTimeout(self.initWait, 1);
```
claygriffiths, line 138: Did you try any JS actions such as [`gpnf_init_nested_form`](https://github.com/gravitywiz/gp-nested-forms/blob/master/js/gp-nested-forms.js#L595)? That's always preferrable over the `setTimeout` / `setInterval` trick 😃

eihabi: @claygriffiths I have, but unfortunately `gpnf_init_nested_form` runs when the modal is deployed now and not when the KO entries are ready. This will cause the population hooks to only work when the child form is displayed.

My only concern with that approach is if there are already pre-populated child entries.

Also agreed, not a fan of `setTimeout` and counters, but there wasn't a readily available hook that I could think of here.

Thoughts?

spivurno: This snippet explicity tries to initialize itself after Nested Forms.
https://github.com/gravitywiz/snippet-library/blob/ab631a392dcc5f7d7418b317c0d591be8dde2493/gp-populate-anything/gppa-populate-child-entries.php#L53

Looks like this is caused by wrapping the self.init() in the jQuery wrapper here: 
https://github.com/gravitywiz/gp-nested-forms/blob/master/js/gp-nested-forms.js#L867-L873

We _could_ solve this by wrapping this init in the same but I don't love that solution as it creates confusion around the loading order. I think the best solution would be to actually refactor that code change in Nested Forms to not require initializing later. There's a chance it might not be needed any more after subsequent changes were made to improve the stability of the performance enhancements. If there are problems removing it, let's discuss and see what can be done.

eihabi: @spivurno I'm a bit warry of refactoring GPNF right now, I'd much rather make the minor modification to this snippet as you suggested (works well, thanks for the hint!). I think we can revisit this issue in a GPNF specific PR. Thoughts?

spivurno: @eihabi I know GPNF is the harder solution but I think it's the right one. We're going to get ourselves in a weird place if we implement a chain of weird solutions across different resources. If you're plate is full, I can take this one.

eihabi: @spivurno I'm totally on board, this just seemed like the path of least resistance 😅 

I might be able to tackle it by EOD tomorrow (if things don't get too crazy). But feel free to take this one if you want 😄 🙏

spivurno: @eihabi All good. I will take this one. 🙂

PR #30, opened 2021-02-23
Add GW Gravity Forms Populate Form Snippet

eihabi:
This PR imports the GW GFPF snippet to the library.

It also adds a new option to allow users to override the snippets population and instead rely on GF's dynamic population feature.

For example, if a user of this snippet uses the URL parameter to import an entry (`eid=xxx`), they can now also use other parameters (`eid=xxx&arg1=yyy&arg2=zzz`) to fill configured fields in the form atop the entry's data.

#22783

### DIFF
--- a/gravity-forms/gw-gravity-forms-populate-form.php
+++ b/gravity-forms/gw-gravity-forms-populate-form.php
@@ -1,0 +1,175 @@
+<?php
+/**
+ * Gravity Wiz // Gravity Forms // Populate Form with Entry (Optionally Update Entry on Submission)
+ *
+ * Pass an entry ID and populate the form automatically. No form configuration required. Optionally update the entry on
+ * submission.
+ *
+ * @version  1.5
+ * @author   David Smith <david@gravitywiz.com>
+ * @license  GPL-2.0+
+ * @link     http://gravitywiz.com/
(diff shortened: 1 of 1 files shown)


gravity-forms/gw-gravity-forms-populate-form.php
```diff
@@ -75,8 +76,15 @@ public function prepare_form_for_population( $form ) {
 				$field['inputs'] = $inputs;
 			}
 
-			$field['inputName'] = $field['id'];
-
+			// If dynamic override is set, only map the field ID if it's empty
+			// Otherwise, leave it as is to populate from GF's dynamic population
+			if ( $this->_args['dynamic_override'] ) {
+				if ( strlen( $field['inputName'] ) < 1 ) {
```
claygriffiths, line 82: Curious, why not `empty` instead of `strlen`?

claygriffiths (approved):
LGTM! Just one non-blocking question

gravity-forms/gw-gravity-forms-populate-form.php
eihabi: @claygriffiths Old habit.  `empty('0')` is truthy. While I don't necessarily see people using `?param1=x&0=y&1=2` in the URL it's nice to know that they can 😅

PR #35, opened 2021-03-10
Fixed a compatibility issue with GF 2.5.

eihabi:
This PR fixes an issue with GF 2.5 and the Custom JS snippet.

GF 2.5 seems to fire the `gform_form_settings` hook before `save_custom_js_setting`, causing this snippet to display an outdated version of the JS code and requiring a page refresh before showing the latest version.

The commit here tweaks `add_custom_js_setting()` to favor the posted value if it's available and differs from the saved version.

#23065

### DIFF
--- a/gravity-forms/gw-gravity-forms-custom-js.php
+++ b/gravity-forms/gw-gravity-forms-custom-js.php
@@ -4,7 +4,7 @@
  *
  * Include custom Javascript with your form.
  *
- * @version  1.5
+ * @version  1.6
  * @author   David Smith <david@gravitywiz.com>
  * @license  GPL-2.0+
  * @link     http://gravitywiz.com/
@@ -13,7 +13,7 @@
  * Plugin URI:   http://gravitywiz.com/
  * Description:  Include custom Javascript with your form.
  * Author:       Gravity Wiz
- * Version:      1.5
+ * Version:      1.6
  * Author URI:   http://gravitywiz.com
  *
  * Usage:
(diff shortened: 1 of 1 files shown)


gravity-forms/gw-gravity-forms-custom-js.php
```diff
@@ -91,12 +91,18 @@ public function noconflict_styles( $scripts = array() ) {
 
 	public function add_custom_js_setting( $settings, $form ) {
 
+		// GF 2.5 may fire `gform_form_settings` before `save_custom_js_setting`
+		$customJS = rgar( $form, 'customJS' );
+		$postJS   = esc_html( rgpost( 'custom_js' ) );
```
spivurno, line 96: These variables should be `$custom_js` and `$post_js` respectively (see our [PHP naming conventions](https://developer.wordpress.org/coding-standards/wordpress-coding-standards/php/#naming-conventions)).

gravity-forms/gw-gravity-forms-custom-js.php
```diff
@@ -91,12 +91,18 @@ public function noconflict_styles( $scripts = array() ) {
...
+		if ( $postJS && $postJS !== $customJS ) {
+			$customJS = $postJS;
+		}
```
spivurno, line 99: I suspect we can simplify the logic here by always using the posted JS if it's present. Was there a scenario you encountered where it didn't make sense to use the posted JS?

spivurno (changes requested):
Good stuff! Just a few minor changes.

gravity-forms/gw-gravity-forms-custom-js.php
eihabi: Ah yes, great catch! It was a JS heavy day 😅

gravity-forms/gw-gravity-forms-custom-js.php
eihabi: That makes sense. Posted JS should always be used since it'll be either the latest version or the current version as is. Fixed :)

spivurno (approved):
Sexiness!

PR #37, opened 2021-03-10
Fixed a potential HTML markup issue.

eihabi:
This PR removes a rouge `</div>` that causes display issues if the snippet is used on a page that contains multiple forms.

#23109


spivurno (approved):
✅

PR #40, opened 2021-03-17
Fixed parsing of `title` and `description` boolean values in shortcodes.

eihabi:
This PR fixes the parsing of the `title` and `description` attributes in shortcodes.

Passing `false` as a string to Gravity Forms' `GFFormDisplay::get_form()` ends up being evaluated as `true` since there's no type checking involved.

The commit in this PR explicitly uses `filter_var` with `FILTER_VALIDATE_BOOLEAN` to correct for that. It also adds `true` as the default setting to match GF's signature.

#23249


claygriffiths (approved):
Interesting usage of `filter_var`!

PR #46, opened 2021-03-24
Gravity Wiz // Gravity Forms // Validate Current Password

sbassah:
Original Gist: https://gist.github.com/spivurno/eebad70856ec3be400ccdeface37479a

### DIFF
--- a/gravity-forms/gw-validate-current-password.php
+++ b/gravity-forms/gw-validate-current-password.php
@@ -6,9 +6,11 @@
 // update "123" to your form ID, update "4" to your Password field ID
 add_filter( 'gform_field_validation_123_4', function( $result, $value ) {
 	$user = wp_get_current_user();
-	if ( ! wp_check_password( $value, $user->data->user_pass, $user->ID ) ) {
-		$result['is_valid'] = false;
-		$result['message'] = 'Invalid current password. Please try again.';
+	if ($user){
+		if ( ! wp_check_password( $value, $user->data->user_pass, $user->ID ) ) {
+			$result['is_valid'] = false;
+			$result['message'] = 'Invalid current password. Please try again.';
+		}
 	}
 	return $result;
 }, 10, 2 );
--- a/gravity-forms/gw-validate-current-password.php
+++ b/gravity-forms/gw-validate-current-password.php
@@ -3,8 +3,8 @@
  * Gravity Wiz // Gravity Forms // Validate Current Password
  * http://gravitywiz.com/
  */
-// update "123" to your form ID, update "6" to your Password field ID
-add_filter( 'gform_field_validation_1639_1', function( $result, $value ) {
+// update "123" to your form ID, update "4" to your Password field ID
+add_filter( 'gform_field_validation_123_4', function( $result, $value ) {
 	$user = wp_get_current_user();
 	if ( ! wp_check_password( $value, $user->data->user_pass, $user->ID ) ) {
 		$result['is_valid'] = false;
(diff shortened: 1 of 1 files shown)


gravity-forms/gw-validate-current-password.php
```diff
@@ -0,0 +1,14 @@
+<?php
+/**
+ * Gravity Wiz // Gravity Forms // Validate Current Password
+ * http://gravitywiz.com/
+ */
+// update "123" to your form ID, update "6" to your Password field ID
```
spivurno, line 6: Update this line to read:
```
// Update "123" to your form ID and "4" to your Password field ID.
```

gravity-forms/gw-validate-current-password.php
```diff
@@ -0,0 +1,14 @@
...
+add_filter( 'gform_field_validation_1639_1', function( $result, $value ) {
```
spivurno, line 7: Update "1639" here to "123" and "1" to "4".

gravity-forms/gw-validate-current-password.php
```diff
@@ -0,0 +1,14 @@
...
+	$user = wp_get_current_user();
+	if ( ! wp_check_password( $value, $user->data->user_pass, $user->ID ) ) {
```
spivurno, line 9: Add a check to see if there the the `$user` variable is truthy.

spivurno (changes requested):

gravity-forms/gw-validate-current-password.php
sbassah: Done

gravity-forms/gw-validate-current-password.php
sbassah: Done

gravity-forms/gw-validate-current-password.php
sbassah: @spivurno I've added this check

gravity-forms/gw-validate-current-password.php
```diff
@@ -0,0 +1,16 @@
...
+ */
+// update "123" to your form ID, update "4" to your Password field ID
+add_filter( 'gform_field_validation_123_4', function( $result, $value ) {
+	$user = wp_get_current_user();
+	if ($user){
+		if ( ! wp_check_password( $value, $user->data->user_pass, $user->ID ) ) {
+			$result['is_valid'] = false;
+			$result['message'] = 'Invalid current password. Please try again.';
+		}
+	}
```
spivurno, line 14: Let's simplify this to a single line check.
```
if ( ! $user || ! wp_check_password( $value, $user->data->user_pass, $user->ID ) ) { 
```

spivurno (changes requested):

gravity-forms/gw-validate-current-password.php
spivurno: I'm not seeing this change?

spivurno: Oh, it looks like you changed the IDs but not the punctutation. You can just copy and paste the exact string I provided. 👌

sbassah: Sorry about that. Guess I was focused on the IDs

gravity-forms/gw-validate-current-password.php
sbassah: I just updated it to a single line check.

spivurno: Haha, don't hate me but I gave you some bad logic. 😆

This should only happen if the user is logged in and the password is bad.

```
if ( $user && ! wp_check_password( $value, $user->data->user_pass, $user->ID ) ) { 
```

spivurno (approved):
One change in the logic (my bad!) and this is good to go!

gravity-forms/gw-validate-current-password.php
sbassah: Yes, you're right. That fixes it.

PR #50, opened 2021-03-24
Gravity Wiz // Enable Decimal Values for Quantity Fields

sbassah:
Original Gist: https://gist.github.com/spivurno/3710739

### DIFF
--- a/gravity-forms/gw-quantity-as-decimal.php
+++ b/gravity-forms/gw-quantity-as-decimal.php
@@ -1,5 +1,9 @@
 <?php
 /**
+ * Gravity Wiz // Gravity Forms // Accept Decimal Values for Quantity Fields
+ *
+ * Allows you to accept decimal values in Quantity fields, converting any Quantity field into a weight field.
+ *
  * Plugin Name:  Gravity Wiz // Gravity Forms // Accept Decimal Values for Quantity Fields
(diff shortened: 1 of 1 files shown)


gravity-forms/gw-quantity-as-decimal.php
```diff
@@ -0,0 +1,128 @@
+<?php
+/**
+ * Gravity Wiz // Gravity Forms // Accept Decimal Values for Quantity Fields
+ *
+ * Allows you to accept decimal values in Quantity fields, converting any Quantity field into a weight field.
+ *
+ * @version   1.3
+ * @author    David Smith <david@gravitywiz.com>
+ * @license   GPL-2.0+
+ * @link      http://gravitywiz.com/enable-decimal-values-in-quantity-fields/
```
spivurno, line 10: Can you replace this with a plugin header?

spivurno (changes requested):

gravity-forms/gw-quantity-as-decimal.php
sbassah: @spivurno  I have updated the snippet with a plugin header.

spivurno: Ah, my intent was only to replace the lines that started with `@` that have been highlighted above. Could you put the other stuff back?

spivurno (approved):
Good work, my friend!

PR #62, opened 2021-03-29
Displays all entries of Nested Forms in a table format.

sbassah:
Original Gist: https://gist.github.com/spivurno/3a3b92bf32824a414c87c23bbf5ca335#file-z-gp-nested-forms-display-child-entries-table-format-all-child-fields-php

### DIFF
--- a/gp-nested-forms/gpnf-display-child-entries-table-format-all-child-fields.php
+++ b/gp-nested-forms/gpnf-display-child-entries-table-format-all-child-fields.php
@@ -1,0 +1,62 @@
+<?php
+/**
+ * Gravity Perks // GP Nested Forms // Display Table Format for All Fields
+ * http://gravitywiz.com/documentation/gravity-forms-nested-forms/
+ * MODIFIED: Show *all* child fields.
+ *
+ * Gravity Perks // GP Nested Forms // Display Table Format for All Fields
+ * Plugin URI:   http://gravitywiz.com/documentation/gravity-forms-nested-forms/
+ * Description:  This snippet displays the Child entries of all Nested Forms field on a form in a table format.
+ * Author:       Gravity Wiz
+ * Version:      0.1
+ * Author URI:   https://gravitywiz.com
+ */
+add_action( 'init', function () {
+
+	if ( !is_callable( 'gp_nested_forms' ) ) {
+		return;
+	}
+
+	add_filter( 'gform_merge_tag_filter', function ( $value, $merge_tag, $modifiers, $field, $raw_value ) {
+
+		if ( $field->type != 'form' || $value === false ) {
+			return $value;
+		}
+
+		$nested_form = GFAPI::get_form( rgar( $field, 'gpnfForm' ) );
+
+		// Adds support for :filter modifier on Nested Form field merge tags but not {all_fields}.
+		$nested_field_ids = wp_list_pluck( $nested_form['fields'], 'id' );
+		if ( function_exists( 'gw_all_fields_template' ) ) {
+			$_modifiers = gw_all_fields_template()->parse_modifiers( $modifiers );
+			if ( $_modifiers['filter'] ) {
+				$nested_field_ids = is_array( $_modifiers['filter'] ) ? $_modifiers['filter'] : array( $_modifiers['filter'] );
+			}
+		}
+
+		$template    = new GP_Template( gp_nested_forms() );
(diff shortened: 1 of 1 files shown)


gp-nested-forms/gpnf-display-child-entries-table-format-all-child-fields.php
```diff
@@ -0,0 +1,63 @@
+<?php
+/**
+ * Gravity Perks // GP Nested Forms // Display Table Format for All Fields
+ * https://gravitywiz.com/documentation/gravity-forms-nested-forms/
+ *
+ * MODIFIED: Show *all* child fields.
```
spivurno, line 6: We should include the full description from the [original snippet](https://github.com/gravitywiz/snippet-library/pull/61/files) and modify it as necessary for this version.

spivurno (approved):
Approved pending new description.

gp-nested-forms/gpnf-display-child-entries-table-format-all-child-fields.php
```diff
@@ -0,0 +1,63 @@
...
+
+	if ( !is_callable( 'gp_nested_forms' ) ) {
+		return;
+	}
+
+	add_filter( 'gform_merge_tag_filter', function ( $value, $merge_tag, $modifiers, $field, $raw_value ) {
+
+		if ( $field->type != 'form' || $value === false ) {
+			return $value;
+		}
```
spivurno, line 25: Just realized this snippet doesn't support activating this via the `gpnf_table` modifier. Can we add support for that? I believe we just need to replace these highlighted lines with these: 

https://github.com/gravitywiz/snippet-library/pull/61/files#diff-54e2bcf2704f530e900c8e5f416f3468733ff75e0d32c5455cb6575ad0f5f91cR22-R24

spivurno (changes requested):

gp-nested-forms/gpnf-display-child-entries-table-format-all-child-fields.php
```diff
@@ -0,0 +1,63 @@
+<?php
+/**
+ * Gravity Perks // GP Nested Forms // Display Table Format for All Fields
```
spivurno, line 3: Let's update "Display Table Format for All Fields" to "Display Table Format for All Fields and Show All Child Fields".

gp-nested-forms/gpnf-display-child-entries-table-format-all-child-fields.php
sbassah: I've updated the snippet with the highlighted lines, however, this version of the snippet actually works without the gpnf_table modifier. Using the modifier and updating the snippet with the lines you suggested, still displays the child entries in the same table format as when no modifier is used. 

Also, there already is a check for !is_callable( 'gp_nested_forms' ) on line 17, which is also present in the line I just added. I think we may have to delete one, but I wanted to confirm if the highlighted lines are still needed since the modifier doesn't affect the snippet.

gp-nested-forms/gpnf-display-child-entries-table-format-all-child-fields.php
sbassah: I have added the description but I had to modify it. This snippet actually works without the gpnf_table modifier, so I removed that part from the description. Please, is this okay?

gp-nested-forms/gpnf-display-child-entries-table-format-all-child-fields.php
sbassah: Done

spivurno:
@sbassah After digging in a bit more, this feels too confusing to manage as two separate snippets. I think the best path forward is to merge the functionality together into a single snippet and allow users to activate the feature they need via modifiers.

```
// Show Nested Form fields in the table format in the {all_fields} merge tag.
{all_fields:gpnf_table}

// Show all child fields when displaying in the table format.
{all_fields:gpnf_table,gpnf_all_fields}
```

I believe all that is necessary to merge these is to update this line in the original snippet (#61):

```
$nested_field_ids = $field->gpnfFields;
```

to:

```
$nested_field_ids = strpos( $modifiers, 'gpnf_all_fields' ) !== false ? wp_list_pluck( $nested_form['fields'], 'id' ) : $field->gpnfFields;
```

Could you close this PR and make that change in a new PR for the `gp-nested-forms/gpnf-display-child-entries-table-format.php` file? Also, update the snippet description and usage examples.

Let me know if you need any help!

PR #64, opened 2021-03-31
Added support for updating author in gw-update-posts.php

scyt:


spivurno:
Didn't install this locally but it looks like this would set the post content to the value of the specified `author` field. 

Also, it might be worth casting the value to an int just to be safe.

```
$post->post_author = (int) rgar( $entry, $this->_args['author'] );
```

Worth noting, I didn't actually test that so probably worth doing a dry run to make sure it works. 😄

spivurno (changes requested):
Didn't install this locally but it looks like this would set the post content to the value of the specified author field.

Also, it might be worth casting the value to an int just to be safe.

```
$post->post_author = (int) rgar( $entry, $this->_args['author'] );
```

Worth noting, I didn't actually test that so probably worth doing a dry run to make sure it works. 😄

scyt:
Whoops. I must have pushed an undo state. Requested changes made. I also bumped the version number and snippet description comment.

spivurno (approved):
Looks good!

PR #65, opened 2021-04-02
Added support for custom modifier values.

spivurno:
[HS#23585](https://secure.helpscout.net/conversation/1471850947/23585/)

Use the new `gwaft_modifier_value_CUSTOMVALUE` filter to replace your custom value with field IDs dynamically.

### DIFF
--- a/gravity-forms/gw-all-fields-template.php
+++ b/gravity-forms/gw-all-fields-template.php
@@ -8,7 +8,7 @@
  * Plugin URI:   https://gravitywiz.com/gravity-forms-all-fields-template/
  * Description:  Modify the {all_fields} merge tag output via a template file.
  * Author:       Gravity Wiz
- * Version:      0.9.8
+ * Version:      0.9.9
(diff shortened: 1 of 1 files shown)


spivurno:
@sbassah Could you review the code, test the new functionality on your end, and then get this on @scyt's plate to update the article with this new feature + usage example?

Might be worth creating a card for this in Notion like we're doing for new snippets and migrations. I'll let you and Scott decide on that. 🙂

spivurno:
Worth noting that this supports passing multiple custom values.

```
{all_fields:exclude[myCustomList,myOtherList]}
```

scyt:
> Might be worth creating a card for this in Notion like we're doing for new snippets and migrations. I'll let you and Scott decide on that. slightly_smiling_face

@sbassah Let's get a card made in Notion once you're done merging this.

sbassah (approved):
It works just as expected on my end. Perfect.

PR #69, opened 2021-04-05
Add Child Entry on Trigger

sbassah:
Original Gist: https://gist.github.com/spivurno/89974e5030b8c68c578b1e55bdfac9c1


spivurno (approved):
Excellent!

PR #71, opened 2021-04-05
Disable Read Only for Admins

sbassah:
Original Gist: https://gist.github.com/spivurno/c37ee64c1a63300e90fe9db6218e959a#file-gp-read-only-disable-for-admins-php

### DIFF
--- a/gp-read-only/gpro-disable-for-admins.php
+++ b/gp-read-only/gpro-disable-for-admins.php
@@ -1,0 +1,22 @@
+<?php
+/**
+ * Gravity Perks // GP Read Only // Disable Read Only for Admins
+ * https://gravitywiz.com/documentation/gravity-forms-read-only/
+ *
+ * This snippet allows you to disbale the Read Only property on fields for Admin users.
+ *
+ * Plugin Name:  GP Read Only — Disable Read Only for Admins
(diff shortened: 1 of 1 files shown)


gp-read-only/gpro-disable-for-admins.php
```diff
@@ -0,0 +1,22 @@
+<?php
+/**
+ * Gravity Perks // GP Read Only // Disable Read Only for Admins
+ * https://gravitywiz.com/documentation/gravity-forms-read-only/
+ *
+ * This snippet allows you to disbale the Read Only property on fields for Admin users.
```
spivurno, line 6: Typo: disbale → disable. 

Not absolutely necessary here since the _why_ is probably easy for most folks to grasp but if I were to add one it'd be something like:
> Disable the Read Only property on fields for Admin users. This is useful when wanting to allow administrative users to edit fields that non-admins should not be able to edit.

spivurno (approved):
Approve pending description update. ✅

gp-read-only/gpro-disable-for-admins.php
sbassah: Thanks for this. I've gotten a better understanding of how to compose the description.

PR #72, opened 2021-04-05
Added support to show all fields in table format

sbassah:
Updated the snippet to support displaying all Child form fields in table format.
Updated the snippet description and usage example


spivurno (approved):
Good work, my friend! ✅

PR #73, opened 2021-04-05
Added usage example back to gw-inventory.php

scyt:
This PR adds the usage example back to gw-inventory.php. It was previously removed, but after further discussion we decided to add usage examples back to the snippets.


spivurno (approved):
Looks good!

PR #74, opened 2021-04-06
Added gw-notification-events.php

scyt:
Original Gist: https://gist.github.com/spivurno/bd9d8e118f45999fb08c

### DIFF
--- a/gravity-forms/gw-notification-events.php
+++ b/gravity-forms/gw-notification-events.php
@@ -1,0 +1,187 @@
+<?php
+/**
+* Gravity Wiz // Gravity Forms // Notification Events
+*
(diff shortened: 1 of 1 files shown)


spivurno (changes requested):
The usage examples here are worth keeping but I think we should add a simpler one. What are we typically handing this out for?

Also, worth noting, this is way more powerful than I remembered it being. 😆

scyt:
I'll come up with a simpler usage example and push an update. We rarely hand this out as-is, but it's required in order to send notifications for gw-inventory.php, which is turned on in the [default config](https://gravitywiz.com/better-inventory-with-gravity-forms/).

spivurno (approved):
Looks good! Went ahead and ran it through PHPCBF too since we were making so many formatting changes anyways. 😄

PR #76, opened 2021-04-09
Fixed an issue with field set with `No Duplicates` causing update to fail.

eihabi:
This PR fixes an issue with the Populate Form with Entry snippet.

If a field is marked as having `No Duplicates` in its settings, the snippet will not work as `gform_entry_id_pre_save_lead` is fired after validation has been completed.

To fix that, this PR adds a new `validate_field` method to filter `gform_field_validation`. It checks if a field that has failed validation did so due to a duplicate entry error while also ensuring that the snippet is being used to update the entry. If all these conditions are met, it overrides the validation error.

#23712


claygriffiths (approved):
LGTM!

PR #79, opened 2021-04-09
Fixed a compatibility issue with GPPA.

eihabi:
This PR fixes an initialization issue when GPPA is used with GW Cache Buster.

It piggy-backs on `gpnf_init_nested_form` if GPPA is detected in the returned markup, and then fires `gform_post_render` to ensure that GPPA has been fully initialized.

#23661

### DIFF
--- a/gravity-forms/gw-cache-buster.php
+++ b/gravity-forms/gw-cache-buster.php
@@ -162,8 +162,7 @@ class GW_Cache_Buster {
 					}
 					// Initialize GPPA
 					if( response.indexOf('GPPA') > -1 ) {
-						window.gform.doAction('gpnf_init_nested_form', formId);
-						jQuery(document).trigger('gform_post_render', [formId, 1]);
+						window.gform.doAction('gppa_register_form', formId);
 					}
 				} );
 			} ( jQuery ) );
--- a/gravity-forms/gw-cache-buster.php
+++ b/gravity-forms/gw-cache-buster.php
@@ -150,6 +150,7 @@ class GW_Cache_Buster {
 		?>
 		<script type="text/javascript">
 			( function ( $ ) {
+				var formId = '<?php echo $form_id ?>';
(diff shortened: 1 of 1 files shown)


spivurno:
I'm wary of triggering an action like `gpnf_init_nested_form` out of context. I know it'll require changes in GPPA to make `maybeRegisterForm()` available in the global scope but I think it's the best path.

As for triggering `gform_post_render`, I think this might be globally necessary in GF 2.5. since the core trigger will be wrapped in a DOMContentLoaded listener which will won't trigger when included in a page like the old jQuery document ready would. With that said, I'm curious what version of GF you were working in and if you found this was necessary in GF 2.4. as well?

eihabi:
@spivurno I thought about making `maybeRegisterForm()` global or even adding a new action like `gppa_register_forms`. That might actually be a better way to go here since I checked for `gpnf_init_nested_form` usage only in GPPA context not in GPNF 🤦‍♂️ , it may have some undesired side effects.

As for GF, this was all reported/tested on 2.4. Cache Buster includes GPPA's markup after `maybeRegisterForm()` has executed, so the `gform_post_render` that's returned in the XHR has no effect since the form wasn't registered with GPPA.

spivurno:
Roger that!

We just want to make sure that `gform_post_render` is only being triggered once (with a strong preference for Gravity Forms' default trigger) so if we only need it for GPPA, it might be better to call the specific functions GPPA is binding to `gform_post_render` event.

spivurno (changes requested):
Clay handed off the official review to me so let's move forward with my previous feedback. 

To summarize, since the issue is specific to GPPA and we don't have any ready example of other GF/perk functionality breaking, let's scope this so that we're only triggering GPPA functions rather than the generic actions/events GPPA binds to.

Unfortunatley, this will require changes to GPPA to make some of its functionality available in the global scope. Ideal in the long term. Annoying for you now. Sorry, buddy! 😜

spivurno (approved):
Love it! Approved pending the changes in the other PR. ✅

PR #85, opened 2021-04-22
Added GPNF Delay Child Notifications to the library

eihabi:
This PR imports the GPNF Delay Child Notifications for Parent Payment snippet.

It also fixes an issue where notifications were sent twice by filtering on `gform_entry_post_save` and specific parent form IDs.

#23899 / DWC.

### DIFF
--- a/gp-nested-forms/gp-nested-forms-delay-child-notifications-for-parent-payment.php
+++ b/gp-nested-forms/gp-nested-forms-delay-child-notifications-for-parent-payment.php
@@ -30,6 +30,7 @@ class GW_GPNF_Delay_Child_Notifications {
 
 		add_filter( 'gpnf_should_send_notification', array( $this, 'gpnf_should_send_notification' ), 10, 7 );
 		add_action( 'gform_post_payment_completed', array( $this, 'gform_post_payment_completed' ) );
+		// Removing this filter causes the original issue of double notification to occur, see HS#23899 PR #85
 		remove_filter( 'gform_entry_post_save', array( gpnf_notification_processing(), 'maybe_send_child_notifications' ), 11 );
 
 	}
@@ -38,8 +39,6 @@ class GW_GPNF_Delay_Child_Notifications {
 		if ( $context === 'parent' && $this->is_applicable_form( $parent_form['id'] ) ) {
 			$parent_entry             = GFAPI::get_entry( rgar( $entry, 'gpnf_entry_parent' ) );
 			$should_send_notification = in_array( rgar( $parent_entry, 'payment_status' ), array( 'Paid', 'Active' ), true );
-			// Prevent double notifications for non-delayed payment gateways.
-			remove_filter( 'gform_post_payment_completed', array( $this, 'gpnf_should_send_notification' ) );
 		}
 
 		return $should_send_notification;
--- a/gp-nested-forms/gp-nested-forms-delay-child-notifications-for-parent-payment.php
+++ b/gp-nested-forms/gp-nested-forms-delay-child-notifications-for-parent-payment.php
@@ -37,7 +37,7 @@ class GW_GPNF_Delay_Child_Notifications {
 		if ( in_array( rgar( $form, 'id' ), $this->form_ids, true ) ) {
 				return $entry;
 		}
-				return gpnf_notification_processing()->maybe_send_child_notifications( $entry, $form );
+		return gpnf_notification_processing()->maybe_send_child_notifications( $entry, $form );
 	}
 }
 
--- a/gp-nested-forms/gp-nested-forms-delay-child-notifications-for-parent-payment.php
+++ b/gp-nested-forms/gp-nested-forms-delay-child-notifications-for-parent-payment.php
@@ -1,0 +1,47 @@
+<?php
+/**
+ * Gravity Perks // Nested Forms // Delay Child Notifications for Parent Payment
+ * http://gravitywiz.com/documentation/gravity-forms-nested-forms/
+ */
+class GW_GPNF_Delay_Child_Notifications {
+	public $form_ids;
+
+	public function __construct( $args ) {
+		$this->form_ids = $args['form_ids'];
+
+		add_filter( 'gpnf_should_send_notification', function( $should_send_notification, $notification, $context, $parent_form, $nested_form_field, $entry, $child_form ) {
+
+			if ( $context === 'parent' ) {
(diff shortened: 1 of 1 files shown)


gp-nested-forms/gp-nested-forms-delay-child-notifications-for-parent-payment.php
```diff
@@ -0,0 +1,47 @@
...
+
+		remove_filter( 'gform_entry_post_save', array( gpnf_notification_processing(), 'maybe_send_child_notifications' ), 11 );
+		add_filter( 'gform_entry_post_save', array( $this, 'gform_entry_post_save' ), 11, 2 );
+	}
+
+	public function gform_entry_post_save( $entry, $form ) {
+		/**
+		 * Do not send notification for form ID 100 on the normal entry save hook. This form is expected to send
+		 * notifications using the gform_post_payment_completed action.
+		 */
```
claygriffiths, line 36: Can you update this comment accordingly?

claygriffiths (commented):
Overall LGTM, just need to sync up one of the comments from moving it into the class 🙂

spivurno:
Solution looks good! 

I'm assuming the customer must be using multiple payment gateways; some that support delayed payments and some that do not. If the user isn't using a payment gateway with delayed payments this wouldn't be necessary. If that's correct, could you add a comment to explain some of the logic here? 

Only other thing is restructing the code a bit. We have a [Snippet Template](https://gravitywiz.slack.com/archives/C03GMA0N7/p1618589901003200) now that will keep a consistent experience for our customers and for us as we maintain these.

spivurno:
Haha, after I explained how all this worked to Clay, I realized that it _might_ be simpler to actually remove the function bound to the `gform_post_payment_completed` filter after `$should_send_notification` return true. This would also account for the scenario where no payment feed is processed and the notification should still be sent. Admittedly, I'm not sure if that is a real-world scenario.

```
if ( $context === 'parent' ) {
    $parent_entry             = GFAPI::get_entry( rgar( $entry, 'gpnf_entry_parent' ) );
    $should_send_notification = in_array( rgar( $parent_entry, 'payment_status' ), array( 'Paid', 'Active' ), true );
    // Prevent double notifications for non-delayed payment gateways.
    remove_filter( 'gform_post_payment_completed', array( $this, 'whatever_you_decide_to_name_this_func' ) );
}
```

eihabi:
@spivurno That did the trick. I've also formatted the snippet to match the template and tested on the user's site.

gp-nested-forms/gp-nested-forms-delay-child-notifications-for-parent-payment.php
```diff
@@ -0,0 +1,62 @@
...
+		// do version check in the init to make sure if GF is going to be loaded, it is already loaded
+		add_action( 'init', array( $this, 'init' ) );
+
+	}
+
+	public function init() {
+
+		add_filter( 'gpnf_should_send_notification', array( $this, 'gpnf_should_send_notification' ), 10, 7 );
+		add_action( 'gform_post_payment_completed', array( $this, 'gform_post_payment_completed' ) );
+		remove_filter( 'gform_entry_post_save', array( gpnf_notification_processing(), 'maybe_send_child_notifications' ), 11 );
```
spivurno, line 33: We shouldn't need this anymore as we're going to conditionally remove our function bound to the `gform_post_payment_completed` filter to account for the double sending.

gp-nested-forms/gp-nested-forms-delay-child-notifications-for-parent-payment.php
```diff
@@ -0,0 +1,62 @@
...
+	}
+
+	public function gpnf_should_send_notification( $should_send_notification, $notification, $context, $parent_form, $nested_form_field, $entry, $child_form ) {
+
+		if ( $context === 'parent' ) {
+			$parent_entry             = GFAPI::get_entry( rgar( $entry, 'gpnf_entry_parent' ) );
+			$should_send_notification = in_array( rgar( $parent_entry, 'payment_status' ), array( 'Paid', 'Active' ), true );
+		}
+		// Prevent double notifications for non-delayed payment gateways.
+		remove_filter( 'gform_post_payment_completed', array( $this, 'gpnf_should_send_notification' ) );
```
spivurno, line 44: This needs to be inside the above conditional so that it only is removed if the notification is going to be sent already.

spivurno (changes requested):

gp-nested-forms/gp-nested-forms-delay-child-notifications-for-parent-payment.php
eihabi: Thanks for catching that! Missed it by a line 😅

gp-nested-forms/gp-nested-forms-delay-child-notifications-for-parent-payment.php
eihabi: Gotcha, done!

spivurno (changes requested):
1. It doesn't look like the `form_ids`  arg is being used anymore. 
2. Can we update this so each instance only applies to a single form? Rather than providing `form_ids` the user would provide a single `form_id` and create an instance of the class for each form. 

    I know some of these details are a undocumented knowledge. Sorry about that. Writing our snippets per form gives more flexibility to control how it works for each form. Doesn't specifically apply here now but if we expand on this with more options in the future we'll be better poised for that granular control.

eihabi:
@spivurno Done. One thing to note, I had to restore the `gform_entry_post_save` filter removal as that seems to have caused double notifications being sent again:
`remove_filter( 'gform_entry_post_save', array( gpnf_notification_processing(), 'maybe_send_child_notifications' ), 11 );`

spivurno:
Code structure looks good! 

And rats! I was hoping we could avoid having to remove the default functionality. Since we're keeping this:

```php
remove_filter( 'gform_entry_post_save', array( gpnf_notification_processing(), 'maybe_send_child_notifications' ), 11 );
```

Let's add a comment explaining why. _And_ remove this:

```php
// Prevent double notifications for non-delayed payment gateways.
remove_filter( 'gform_post_payment_completed', array( $this, 'gpnf_should_send_notification' ) );
```

It doesn't the work the way I was hoping it would.

eihabi:
@spivurno Done!

spivurno (approved):
Thanks for sticking it out, Eihab! Approved. ✅

PR #86, opened 2021-04-22
Added support for updating post categories and custom taxonomies.

scyt:
This adds support for updating post categories and custom taxonomies. It assumes the user is populating taxonomies into a Checkbox field.

### DIFF
--- a/gravity-forms/gw-update-posts.php
+++ b/gravity-forms/gw-update-posts.php
@@ -17,88 +17,75 @@ class GW_Update_Posts {
 		$this->_args = wp_parse_args(
 			$args,
 			array(
-				'form_id'    => false,
-				'post_id'    => false,
-				'title'      => false,
-				'content'    => false,
-				'author'     => false,
-				'terms'      => false,
-				'meta' => array()
+				'form_id' => false,
+				'post_id' => false,
+				'title'   => false,
+				'content' => false,
+				'author'  => false,
+				'terms'   => false,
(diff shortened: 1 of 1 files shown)


spivurno (approved):
@scyt Good work, man, Particularly clever solution for devising the taxonomy from the populated choice's object. I thought it might be better to get it from the GPPA filters for the field but after digging in, I think your solution is more reliable.

Since you're traveling today, I went ahead and made a few updates so I can send this on to a customer I have waiting on me to approve this.

1. I wrapped each post property update in a check to make sure that the argument was set and passed a value. One of the first issues I ran into in testing was the assumption that I could update just a specific part of the post. This is now supported.
2. I updated the `RGFormsModel::get_field()` call to use the `GFAPI` equivalent instead. Both work but the GFAPI is safer since it's GF's public API and RGFormsModel is technically for internal use only. So much undocumented knowledge. 😜
3. I removed the `return_ids_instead_of_names` method and filter. In my testing it wasn't necessary when Term ID was set as the value which I believe to be a fair requirement from us. Let me know if there was another issue related to this.

scyt:
@spivurno `return_ids_instead_of_names` is necessary when using taxonomies with hierarchies. WP expects an array of term IDs for any taxonomies that are hierarchical, but GPPA returns Term Name when populating terms as values. Without the filter, the selected post's terms won't populate into the checkbox.

PR #89, opened 2021-04-26
Fixed an issue where field values were not passed properly to GF.

eihabi:
This PR fixes an issue where `field_values` were not being passed properly in the short code.

An example cause would look like: `[gravityform id="233" title="false" cachebuster="1" description="false" field_values="webinar_code=123-456-789"]`.

GF expects the `$field_values` parameter to be an associative array. The commit here fixes that by using `parse_str()`.

#24038

### DIFF
--- a/gravity-forms/gw-cache-buster.php
+++ b/gravity-forms/gw-cache-buster.php
@@ -184,9 +184,13 @@ class GW_Cache_Buster {
 			$form_id = isset( $_GET['form_id'] ) ? absint( $_GET['form_id'] ) : 0;
 		}
 
-		$atts = json_decode( rgpost( 'atts' ), true );
-
-		gravity_form( $form_id, filter_var( rgar( $atts, 'title', true ), FILTER_VALIDATE_BOOLEAN ), filter_var( rgar( $atts, 'description', true ), FILTER_VALIDATE_BOOLEAN ), false, rgar( $atts, 'field_values' ), true /* default to true; add support for non-ajax in the future */, rgar( $atts, 'tabindex' ) );
+		$atts         = json_decode( rgpost( 'atts' ), true );
+		$field_values = rgar( $atts, 'field_values' );
+		// GF expects an associative array for field values. Parse them before passing it on.
+		if ( $field_values ) {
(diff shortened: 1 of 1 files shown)


gravity-forms/gw-cache-buster.php
```diff
@@ -184,9 +184,13 @@ public function ajax_get_form() {
...
 		}
 
-		$atts = json_decode( rgpost( 'atts' ), true );
-
-		gravity_form( $form_id, filter_var( rgar( $atts, 'title', true ), FILTER_VALIDATE_BOOLEAN ), filter_var( rgar( $atts, 'description', true ), FILTER_VALIDATE_BOOLEAN ), false, rgar( $atts, 'field_values' ), true /* default to true; add support for non-ajax in the future */, rgar( $atts, 'tabindex' ) );
+		$atts         = json_decode( rgpost( 'atts' ), true );
+		$field_values = rgar( $atts, 'field_values' );
+		// GF expects an associative array for field values. Parse them before passing it on.
+		if ( $field_values ) {
+			parse_str( $field_values, $field_values );
```
claygriffiths, line 191: Can this be replaced with https://developer.wordpress.org/reference/functions/wp_parse_args/?

claygriffiths (approved):
LGTM pending the switch to `wp_parse_args`!

gravity-forms/gw-cache-buster.php
eihabi: @claygriffiths `wp_parse_args()` performs a few checks `is_object()`, `is_array()` and finally ends up using `wp_parse_str()` for strings which ends up using PHP's built-in `parse_str()` and applying a `wp_parse_str` filter.

Do we need this filter to fire or is this direct call acceptable?

claygriffiths: I believe replacing these four lines with...

```php
$field_values = wp_parse_args( rgar( $atts, 'field_values' ) );
```

... should do the trick! Let me know if not.

eihabi: @claygriffiths Done. Merging and sending out.

PR #94, opened 2021-05-04
Add support for updating multiple taxonomies in a single entry.

scyt:
This PR adds support for updating multiple taxonomies with a single form entry. For example, you could update the post's categories and custom taxonomies at the same time.


spivurno (approved):
Looks good!

PR #96, opened 2021-05-07
Disable auto-complete on form or on each field

sbassah:
Original Gist: https://gist.github.com/spivurno/34963fcc649949fa49d3

I didn't add the plugin header because the snippet is short,


spivurno:
@sbassah We've made some improvements to this snippet and already migrated it. Haha, sorry again! 😆

PR #98, opened 2021-05-14
Filter Results on Relative Dates

sbassah:
Original Gist: https://gist.github.com/claygriffiths/bdd06373bb04ab7a62fd0330c23d619d


gp-populate-anything/gppa-filter-by-relative-dates.php
```diff
@@ -0,0 +1,45 @@
+<?php
+/**
+ * Gravity Perks // GP Populate Anything // Filter Results on Relative Dates
```
spivurno, line 3: Remove GP from the snippet name like so:
> Gravity Perks // Populate Anything // Filter Results on Relative Dates

spivurno (approved):
Approved with one minor tweak.

PR #99, opened 2021-05-25
Fixed compatibility issue with GF 2.5 where conditionally shown fields would cause the form to not render.

eihabi:
With GF 2.5, the `gform_post_render` trigger that gets appended with the markup doesn't seem to fire.

This causes forms with conditionally hidden fields to remain hidden (`style="display:none"`) as `conditionallogic.js` never gets notified to process the logic and show the form.

This PR gets around that by manually triggering `gform_post_render` if GF's version is `>=` 2.5.

#24394


spivurno:
Man, GF 2.5 has really introduced some weirdness with init scripts! 😜

Deferring to Clay for a proper review but I'd love to see one little change here. Could you update the comment to explain why rather than what we're doing?

eihabi:
@spivurno Ah, my bad. Updated.

eihabi:
Closing this as it appears to have been fixed in GF 2.5.2.4.

PR #101, opened 2021-06-03
Fixed an issue with GF2.5 where rows were not counted correctly.

eihabi:
Snippet: gw-set-list-field-rows-by-field-value.php

This PR fixes a markup issue with GF 2.5. By default, the new markup doesn't use tables for List fields.

This caused the snippet to add an unlimited number of fields with every interaction as it was not able to get the current number accurately.

With a version check, the updated snippet will now look for `.gfield_list_group` in GF 2.5 with a fallback to the table method for 2.4 and below.

[#24871](https://secure.helpscout.net/conversation/1527600936/24871/)

### DIFF
--- a/gravity-forms/gw-set-list-field-rows-by-field-value.php
+++ b/gravity-forms/gw-set-list-field-rows-by-field-value.php
@@ -16,146 +16,146 @@
 class GWAutoListFieldRows {
 
 	private static $_is_script_output;
-	
+
 	public function __construct( $args = array() ) {
- 
+
 		// set our default arguments, parse against the provided arguments, and store for use throughout the class
 		$this->_args = wp_parse_args( $args, array(
 			'form_id'       => false,
 			'input_html_id' => false,
-			'list_field_id' => false
+			'list_field_id' => false,
 		) );
- 
+
 		// do version check in the init to make sure if GF is going to be loaded, it is already loaded
 		add_action( 'init', array( $this, 'init' ) );
- 
+
 	}
- 
+
 	public function init() {
- 
+
 		// make sure we're running the required minimum version of Gravity Forms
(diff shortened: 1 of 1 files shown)


gravity-forms/gw-set-list-field-rows-by-field-value.php
```diff
@@ -16,146 +16,147 @@
...
+					this.updateListItems = function( elem, listFieldId, formId ) {
 
-                        var listField = $( '#field_' + formId + '_' + listFieldId ),
-                            count = parseInt( elem.val() );
-                        rowCount = listField.find( 'table.gfield_list tbody tr' ).length,
-                            diff = count - rowCount;
+						var listField = $( '#field_' + formId + '_' + listFieldId ),
+							count = parseInt( elem.val() );
+						// GF 2.5 doesn't use tables by default, look for `gfield_list_group` in that case with GF2.4 fallback
+						rowCount = ( gf25 ) ? listField.find('.gfield_list_group').length : listField.find( 'table.gfield_list tbody tr' ).length,
```
claygriffiths, line 92: I believe this can be simplified to just `listField.find('.gfield_list_group').length` which should work with both GF 2.4 and 2.5.

claygriffiths (commented):
One quick simplification.

Additionally, can you include the filename in all of the commits?

gravity-forms/gw-set-list-field-rows-by-field-value.php
eihabi: @claygriffiths Great call! I didn't check GF 2.4 to confirm if that would work there too.

I went ahead and removed the 2.5 check and simplified the rowCount setup.

eihabi:
@claygriffiths 10-4! Will add snippet name in future commits. I went ahead and updated the description on this one for posterity.

claygriffiths (approved):

PR #102, opened 2021-06-03
Initial support for rounding the total field

eihabi:
@claygriffiths This PR is a work in progress. I've hit a logical wall here with adding rounding support to total fields.

Rounding on the front-end is straight forward, I've added a JS hook [`gform_product_total`](https://docs.gravityforms.com/gform_product_total/) which then checks for the proper classes and rounds the total correctly.

The issue now is that this is just cosmetic. We need to implement [`gform_product_info`](https://docs.gravityforms.com/gform_product_info/) on the back-end to actually change the amount users will pay.

The problem here is that when applying the rounding to a total, there's no telling which product(s) prices should be affected.

My initial idea was to just split the rounded total by the quantity on the first product and then add an "adjusted rounded price" like `$product_info['products']['adjusted_rounded_price']` with the difference.

That will not work though if the form has more than one product and will also cause 3rd party processing to fail.

Any thoughts here about what can be done or is this just outside of the scope of what we can support?

[#24946](https://secure.helpscout.net/conversation/1531948301/24946?folderId=3808239)

### DIFF
--- a/gravity-forms/gw-gravity-forms-rounding.php
+++ b/gravity-forms/gw-gravity-forms-rounding.php
@@ -152,8 +152,6 @@ class GW_Rounding {
 
 							var $targets;
 
-							debugger;
-
 							// Attempt to target only the quantity input of a Single Product field.
 							if( $( this ).hasClass( 'gfield_price' ) ) {
 								$targets = $( this ).find( '.ginput_quantity' );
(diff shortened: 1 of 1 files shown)


eihabi:
Per [David's response](https://secure.helpscout.net/conversation/1531948301/24946?folderId=3808239), leaving this open but removing debugger in #103.

PR #104, opened 2021-06-08
Add gw-disable-autocomplete.php to snippet library. Added name attrib…

scyt:
…ute to target Safari.


scyt:
Removed the Safari hack because it broke submission.

spivurno:
@scyt With the removal of the Safari hack, are there any changes to this snippet or just the migration to the library?

spivurno (approved):
Got it. Looks good!

PR #105, opened 2021-06-11
Added support for updating post status.

scyt:
The idea for this is a choice field with the different post status as available choices.

This choice field can be populated via GPPA:

### DIFF
--- a/gravity-forms/gw-update-posts.php
+++ b/gravity-forms/gw-update-posts.php
@@ -67,7 +67,11 @@ class GW_Update_Posts {
 		}
 
 		if ( $this->_args['status'] ) {
-			$post->post_status = rgar( $entry, $this->_args['status'] );
+			$stati = get_post_stati();
+			$status = rgar( $entry, $this->_args['status'] );
+			if ( in_array( $status, $stati ) ) {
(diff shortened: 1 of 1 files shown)


gravity-forms/gw-update-posts.php
```diff
@@ -65,6 +66,10 @@ public function update_post_by_entry( $entry, $form ) {
 			$post->post_author = (int) rgar( $entry, $this->_args['author'] );
 		}
 
+		if ( $this->_args['status'] ) {
+			$post->post_status = rgar( $entry, $this->_args['status'] );
+		}
```
spivurno, line 71: It'd be awesome if we could get a list of [registered post statuses](https://developer.wordpress.org/reference/functions/get_post_stati/) and then make sure that the user-selected (or enetered) status is a valid post status before we attempt to set it. If not, we could just skip updating it.

spivurno (changes requested):
I dig it! Great summary as well. Just one reliability note below.

scyt:
Good call. I added a check to see if the post status exists. If it doesn't, it skips updating.

spivurno (approved):
Perfect!

PR #106, opened 2021-06-11
Fixed an issue where "Order Summary" label would still appear when using `nopricingfields` modifier.

eihabi:
The `nopricingfields` modifier doesn't seem to apply to the "Order Summary" label.

This PR uses a second filter `gform_display_product_summary` to ensure it's hidden.

Ticket: [#25152](https://secure.helpscout.net/conversation/1539696626/25152/)

### DIFF
--- a/gravity-forms/gw-all-fields-template.php
+++ b/gravity-forms/gw-all-fields-template.php
@@ -240,7 +240,7 @@ class GW_All_Fields_Template {
 		// $field_id = $field->id;
 		// print_r( compact( 'modifiers', 'field_ids', 'field_id', 'value' ) );
 		// echo '<pre>';
-
+		remove_filter( 'gform_display_product_summary', '__return_false' );
 		return $value;
 	}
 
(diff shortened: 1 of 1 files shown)


gravity-forms/gw-all-fields-template.php
```diff
@@ -166,6 +166,8 @@ public function all_fields_extra_options( $value, $merge_tag, $modifiers, $field
 					return '';
 				};
 				add_filter( 'gform_order_summary', $func );
+				// Hide "Order Summary" label if nopricingfields is used
+				add_filter( 'gform_display_product_summary', '__return_false' );
```
claygriffiths, line 170: I'm curious if this will cause issues if for some reason you use `{all_fields:nopricingfields}` and  `{all_fields}`  in the same confirmation/notification. 

If so, you may want to remove this filter right before `return $value`.

eihabi: @claygriffiths Tested and fixed by removing filter before returning `$value`.

claygriffiths (approved):

PR #109, opened 2021-06-17
Fixed an issue where `gppa-populate-child-entries.php` may generate a JS error in multi-page forms.

eihabi:
This PR fixes an issue where [gppa-populate-child-entries](https://github.com/gravitywiz/snippet-library/blob/master/gp-populate-anything/gppa-populate-child-entries.php) would generate a JS error in multi-paged forms.

This snippet ([Loom demo by David](https://www.loom.com/share/7b26b18f78624e0ca4bcd2b574636b8b)) assumed that GPNF would have instantiated Knockout.JS on page load. It already used the correct filter `gpnf_session_initialized`, however, the for loop needed to be inside that closure as well.

Ticket: [#25190](https://secure.helpscout.net/conversation/1539872947/25190?folderId=3808239)

### DIFF
--- a/gp-populate-anything/gppa-populate-child-entries.php
+++ b/gp-populate-anything/gppa-populate-child-entries.php
@@ -97,13 +97,14 @@ class GPPA_Populate_Child_Entries {
 									.val( gpnfCookie.hash )
 									.change();
 							}
-						} );
 
-						for ( var i = 0; i < self.nestedFormFieldIds.length; i++ ) {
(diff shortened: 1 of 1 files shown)


spivurno (approved):
Did you confirm if this still worked well on single-page forms? Assuming so, this is approved!

Also to clarify, is the Nested Form on a subsequent page of the multi-page form and therefore not initialized yet?

eihabi:
@spivurno Yes, I have removed paging and tested that the snippet still works!

And yes, the nested form was on page 2 which caused this snippet to fail.

I've tested both scenarios and confirmed it works on both.

Merging/sending 👍

PR #119, opened 2021-06-27
Added support for updating featured image via File Upload field.

spivurno:
[HS#25449](https://secure.helpscout.net/conversation/1551081560/25449/)


scyt (approved):
Works perfectly! 👍

PR #122, opened 2021-07-04
Added support for randomizing field order.

spivurno:
[HS#25521](https://secure.helpscout.net/conversation/1553478273/25521)

This PR adds support for randomizing field order which is enabled by default. This feature can be disabled by passing `true` for the new `$perserve_field` parameter when instantiating the class.

`new GFRandomFields( 123, 2, array( 1, 2, 4, 5, 7 ), true );`


spivurno:
@scyt Could you review for functionality? I'll send over to Eihab for a code review once you're done. 🙂

scyt (approved):
Works great!

eihabi (approved):
@spivurno LGTM!

PR #125, opened 2021-07-22
Fixed an issue where `gw-create-coupon` scheduled coupons using GMT instead of the timezone configured in WordPress.

eihabi:
This PR fixes an issue with the `gw-create-coupon` snippet's handling of start date with WooCommerce.

The snippet is supposed to set the coupons to start immediately, however if the timezone is behind GMT (e.g. `America/New_York`) the coupon gets scheduled to a time in the future and is not available for immediate use.

The fix here was to swap `gmdate()` with `current_datetime()` to respect the timezone set in WP.

Ticket: [#26099](https://secure.helpscout.net/conversation/1577829998/26099?folderId=3808239)

### DIFF
--- a/gravity-forms/gw-create-coupon.php
+++ b/gravity-forms/gw-create-coupon.php
@@ -185,7 +185,8 @@ class GW_Create_Coupon {
 
 		$start_date = rgar( $this->_args['meta'], 'start_date' );
 		if ( $start_date === '' || ! strtotime( $start_date ) ) {
-			$start_date = gmdate( 'Y-m-d H:i:s' );
+			$date       = current_datetime();
+			$start_date = $date->format( 'Y-m-d H:i:s' );
 		}
 
 		// WooCommerce coupon uses the Post Title as the coupon code hence $coupon_code is assigned to Post Title and $coupon_name is assigned to the Post Content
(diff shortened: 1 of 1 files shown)


gravity-forms/gw-create-coupon.php
```diff
@@ -185,7 +185,8 @@ public function create_coupon_wc( $coupon_name, $coupon_code, $amount, $type, $e
 
 		$start_date = rgar( $this->_args['meta'], 'start_date' );
 		if ( $start_date === '' || ! strtotime( $start_date ) ) {
-			$start_date = gmdate( 'Y-m-d H:i:s' );
+			$date       = current_datetime();
+			$start_date = $date->format( 'Y-m-d H:i:s' );
```
claygriffiths, line 189: I think this can be simplified to changing `post_date` on L198/199 to `post_date_gmt`

eihabi: @claygriffiths My apologies, I didn't clarify in the PR. This case only applies if the end user doesn't explicitly set a `start_date`. Modifying L198 would make all coupons available immediately and remove scheduling functionality.

claygriffiths: Ahh, good point! ✅

claygriffiths (approved):

PR #127, opened 2021-07-24
Added support for removing featured image if no value is submitted.

spivurno:
**[HS#26075](https://secure.helpscout.net/conversation/1576510052/26075/)**

This PR adds supports for removing the selected post's featured image if no value is submitted for featured image field specified in the configuration.

My primary concern with this is that we currently have no way to repopulate the current featured image from the post. It might be better to make this an opt-in feature with an additional configuration parameter. Thoughts?

### DIFF
--- a/gravity-forms/gw-update-posts.php
+++ b/gravity-forms/gw-update-posts.php
@@ -17,15 +17,17 @@ class GW_Update_Posts {
 		$this->_args = wp_parse_args(
 			$args,
 			array(
-				'form_id'        => false,
-				'post_id'        => false,
-				'title'          => false,
-				'content'        => false,
-				'author'         => false,
-				'status'         => false,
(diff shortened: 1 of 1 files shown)


eihabi (approved):
@spivurno LGTM! I'm not sure about the optional parameter here, so I'm deferring to @claygriffiths for final approval.

claygriffiths (commented):
Yeah, considering it won't repopulate, I'd lean toward making this use an option; at the very least, be disable-able.

spivurno:
@claygriffiths Roge that. Added a new parameter: "delete_if_empty". Defaults to `false`. Set to `true` to delete featured images when the parameter is mapped but no value is submitted. We can expand this to support other parameters in the future based on customer demand.

claygriffiths (approved):

PR #128, opened 2021-07-27
Added support for requiring submission on current page with Snippet to Access using `per_page` value in `gwsa_require_submission` meta

claygriffiths:
## Summary

Add option to use `per_page` as value for `gwsa_require_submission` meta to require forms to be submitted on the current page.

This allows you to re-use a form on multiple pages while requiring it to be resubmitted each time.

### DIFF
--- a/gravity-forms/gw-submit-to-access.php
+++ b/gravity-forms/gw-submit-to-access.php
@@ -266,11 +266,12 @@ class GW_Submit_Access {
 		}
 
 		$form_ids = $this->get_form_ids( $post_id );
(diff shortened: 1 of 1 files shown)


spivurno (approved):
Flawless.

Tested with two pages and the following variations.

- `per_page` enabled for Page A and Page B. 

    Required submission of the specified form on each page to view the given page's content.

- `per_page` enabled for Page A and Page B used default enabled mode.

    - Page content was viewable on both pages if form on Page A was submitted.
    - Page content was viewable only on Page B if Page B was submitted. Page A content was viewable only after form was submitted from Page A.

PR #129, opened 2021-07-27
Fixed an issue with GF 2.5 Checkbox fields and `gw-cache-buster`.

eihabi:
GF 2.5 expects `$_POST` to be an empty `Array` before applying defaults to Checkbox fields.

While this is a Gravity Forms core issue, emptying out `$_POST` in GWCB resolves this issue.

I've modified @claygriffiths's patch slightly by not retaining `$_POST` in a variable. We `die()` immediately after GF outputs the form's markup.

Ticket: [#26188](https://secure.helpscout.net/conversation/1580050306/26188/)


claygriffiths (approved):
Thanks for adding this! Can you leave a quick inline comment on the why?

PR #130, opened 2021-08-04
Improved `delete_if_empty` argument by adding support for custom fields.

eihabi:
This PR expands #127 by deleting custom fields if the source is empty and `delete_if_empty` option is specified.

Ticket: [#26286](https://secure.helpscout.net/conversation/1584354402/26286?folderId=3808239)


claygriffiths (approved):
Beautimous!

PR #133, opened 2021-08-13
Added support for populating/modifying times from one Time field to another.

spivurno:
[HS#26609](https://secure.helpscout.net/conversation/1599101905/26609)

This PR adds support for populating/modifying times from one Time field into another. 

Here is a form export:

[gravityforms-export-2021-08-13.json.zip](https://github.com/gravitywiz/snippet-library/files/6984892/gravityforms-export-2021-08-13.json.zip)

...and example configuration:

```php
new GW_Populate_Date( array(
	'form_id'         => 123,
	'source_field_id' => 1,
	'target_field_id' => 2,
	'modifier'        => array(
		'type'     => 'field',
		'inputId'  => 4,
		'modifier' => '+{0} minutes',
	),
) );
```

### DIFF
--- a/gravity-forms/gw-populate-date.php
+++ b/gravity-forms/gw-populate-date.php
@@ -143,24 +143,43 @@ class GW_Populate_Date {
 			$raw = array_filter( $raw );
 		}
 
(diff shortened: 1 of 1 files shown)


claygriffiths (approved):
LGTM and tested locally.

This is a beefy snippet! 😱

PR #135, opened 2021-08-23
Added support to Update Post Name (Slug)

sbassah:
Added support to update the Post slug using a Field Value. The parameter name is 'slug' and the user will be required to enter the ID of the field.


spivurno (approved):
Great work, Samuel!

PR #137, opened 2021-08-30
Fix conflicts with other plugins

shaunjc:
This plugin causes a conflict with other plugins and themes that use the `gform_merge_tag_filter` and require all 6 arguments, however this plugin does not supply the `$format` parameter.

    function my_gform_merge_tag_filter( $value, $merge_tag, $modifier, $field, $raw_value, $format ) {
        // Filter based on $format.
        return $value;
    }
    add_filter( 'gform_merge_tag_filter', 'my_gform_merge_tag_filter', 10, 6 );
    // PHP Fatal error:  Uncaught ArgumentCountError: Too few arguments to function my_gform_merge_tag_filter

This should be a relatively simple change, by supplying that the format is "html".

### DIFF
--- a/gravity-forms/gw-all-fields-template.php
+++ b/gravity-forms/gw-all-fields-template.php
@@ -433,7 +433,7 @@ class GW_All_Fields_Template {
 						$field_value = false;
 					}
 
-					$field_value = apply_filters( 'gform_merge_tag_filter', $field_value, $merge_tag, $modifiers, $field, $raw_field_value );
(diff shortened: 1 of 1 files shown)


claygriffiths:
Hi @shaunjc, thanks for the PR! It looks like @eihabi has a similar PR (https://github.com/gravitywiz/snippet-library/pull/120) that addresses the same issue. We'll get it merged ASAP.

PR #141, opened 2021-09-02
Use List Field as Choices for Gravity Forms

sbassah:
Original Gist: https://gist.github.com/spivurno/8202318#file-gw-gravity-forms-list-field-as-choices-php

### DIFF
--- a/gravity-forms/gw-list-field-as-choices.php
+++ b/gravity-forms/gw-list-field-as-choices.php
@@ -1,0 +1,145 @@
+<?php
+/**
+* Gravity Wiz // Gravity Forms // Use List Field as Choices for Gravity Forms
+* https://gravitywiz.com/use-list-field-choices-gravity-forms/
+*
+* Adds support for populating choice-based fields (i.e. checkboxes, selects, radio buttons) with values entered in a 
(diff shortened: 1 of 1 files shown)


gravity-forms/gw-list-field-as-choices.php
```diff
@@ -0,0 +1,145 @@
+<?php
+/**
+* Gravity Wiz // Gravity Forms // Use List Field as Choices for Gravity Forms
+* https://gravitywiz.com/use-list-field-choices-gravity-forms/
+*
+* Adds support for populating choice-based fields (i.e. checkboxes, selects, radio buttons) with values entered in a 
+* List field. This functionality requires that the form has multiple pages and that the List field must be placed on 
+* a page prior to the choice-based field that it will populate.
+*
+* Plugin Name:  Gravity Forms - Use List Field as Choices for Gravity Forms
```
spivurno, line 10: Let's drop the second "Gravity Forms" from the plugin name. Should look like this:
> Gravity Forms - Use List Field as Choices

spivurno (approved):
Approved with one change. 🙂

PR #143, opened 2021-09-03
Added support for filtering the displayed states based on the allowed states.

spivurno:
Bit of a hack but I think it makes the snippet a lot more practical! Thoughts? 

**Edit:** This also fixes an issue where all Address fields were validated even when a `field_id` was specified.


claygriffiths (approved):
Nice!

✅ pending a `@version` bump on the snippet

PR #145, opened 2021-09-09
Move Currency Symbol from the Right to the Left

sbassah:
Original Gist: https://gist.github.com/spivurno/0d772bae83bb216e89aa


spivurno (changes requested):
Can we update this to use an anonymous function? I think it will simplify the code.

sbassah:
I have made the requested changes.

spivurno (approved):
Thanks, man!

PR #150, opened 2021-09-19
Refactored custom JavaScript to be enqueued as an init script for better compatibility with Gravity Forms 2.5.

spivurno:
[HS#27390](https://secure.helpscout.net/conversation/1630758040/27390)

This PR refactors custom JS to be enqueued as an init script rather than output directly to the footer. This resolves an issue with Gravity Forms 2.5 where custom JS (which is bound by default to the `gform_post_render` event) was being triggered before registered init scripts instead of after (as is the case with GF 2.4).

By registering the custom JS as an init script on a very late priority (e.g. 99), we can ensure that it is triggered after all other init scripts.

I was unable to detect a way to do this without registering the JS as an init script. There is no window where the script could be bound to the `gform_post_render` event _after_ all other scripts without missing the actual triggering of the `gform_post_render` event. This is due to GF 2.5 using the new `gform.gform.initializeOnLoaded()` and the removal of use of the jQuery document ready event.

In the case of this ticket, this change allows us to be sure that Nested Forms has fully initialized before we attempt to subscribe to child entry changes. See the experimental snippet here:

https://github.com/gravitywiz/snippet-library/blob/master/experimental/gpnf-count-answered-child-fields.js

### Follow-up

**New Feature: GFORMID**

I didn't mention this before but I also added a new "constant": `GFFORMID` that will automatically be replaced on render with the current form ID. This allows us to create more generic JS snippets that users can just copy and paste into our Custom Javascript plugin and not have to edit the form ID match their own.

**Improved Compat w/ GF 2.5**

The settings now display in their own native GF 2.5 section instead of the "Legacy Settings" section.

I would _love_ to explore adding tabs here to let users specify custom CSS as well. Leagues better than putting it in an HTML field.

### DIFF
--- a/gravity-forms/gw-gravity-forms-custom-js.php
+++ b/gravity-forms/gw-gravity-forms-custom-js.php
@@ -57,7 +57,7 @@ class GF_Custom_JS {
 			add_filter( 'gform_noconflict_styles', array( $this, 'noconflict_styles' ) );
 		}
 
-		add_filter( 'gform_pre_render', array( $this, 'load_form_script' ), 10, 2 );
+		add_filter( 'gform_register_init_scripts', array( $this, 'register_init_script' ), 99 );
 
 	}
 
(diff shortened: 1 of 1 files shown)


claygriffiths (approved):
Very nice work! All great improvements.

PR #151, opened 2021-09-20
Fixed issue where first field with error was scrolled to but not focused.

spivurno:
[HS#27092](https://secure.helpscout.net/conversation/1618555630/27092)

This PR resolves an issue where the first field error was scrolled to but not given focus. The issue was introduced in 2e13614229491dee355ce8eb3406c7fda0c57a3b where scrolling was delayed but the focus was not. A field that is not in the viewport cannot be focused (I guess?).

To resolve this, we simply set the focus after the element has been scrolled to.


spivurno:
@scyt Could you give this a functional review?

claygriffiths (approved):

scyt:
@spivurno Looks good to me!

PR #156, opened 2021-09-23
Fixed an issue where `gw-populate-date` over-wrote populated date on form submission.

eihabi:
This PR fixes an issue where a date field that's been modified by the snippet would revert back to the calculated date on submission.

It looks like the option `override_on_submission` was added before, but it was not being applied when a `source_field_id` existed.

Ticket: [#27634](https://secure.helpscout.net/conversation/1639770700/27634?folderId=3808239)

### DIFF
--- a/gravity-forms/gw-populate-date.php
+++ b/gravity-forms/gw-populate-date.php
@@ -46,7 +46,9 @@ class GW_Populate_Date {
 		}
 
 		if ( $this->_args['source_field_id'] ) {
-			add_action( 'gform_pre_submission', array( $this, 'populate_date_on_pre_submission' ) );
+			if ( $this->_args['override_on_submission'] ) {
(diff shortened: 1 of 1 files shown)


spivurno (changes requested):
This change makes sense since we added support for automatically populating the target field from the source field on render (via JS). 

The only strange thing is having two methods that are performing very similar functions.

`populate_date_on_pre_submission()`
`override_on_submission()`

Could you take a look at consolidating them as part of this PR? Let me know if you want to have a quick pow wow to answer any questions on original intent. 😄

eihabi:
@spivurno Done, consolidated both methods into `populate_date_on_pre_submission()` and added the source field distinction there :+1:

spivurno (approved):
@eihabi Most excellent!

PR #158, opened 2021-09-27
gw-submit-to-access use default message

ethanclevenger91:
Default submission message should be added before appending the form markup, or it will never be "empty".


spivurno (approved):
Thanks, @ethanclevenger91! Good catch. 🙂

PR #159, opened 2021-09-30
Update to svg

ethanclevenger91:
Gravity Forms swapped their GIF for an SVG


spivurno:
Most excellent, @ethanclevenger91! 

We're maintaining backwards compatibility with GF 2.4. Could you add a version check here to ensure that the `GFForms::$version` is `2.5` or greater and if so, use the SVG?

PR #162, opened 2021-10-01
Added support for using the `:img` modifier with fields populated with URLs rather than attachment IDs as well.

spivurno:
[HS#27799](https://secure.helpscout.net/conversation/1646377681/27799)

This PR updates GPPA's `:img` modifier snippet to support outputting raw URLs as images in addition to the existing functionality of outputting images for attachment ID values.

The field's value can be a single URL or a comma-delimited list of URLs. This is relevant when populating values from Single and Multi-file Upload fields into a text-based field (e.g. Single Line Text).


scyt (approved):
LGTM!

PR #163, opened 2021-10-05
Auto-attach Uploaded Files to Parent Notifications

sbassah:


gp-nested-forms/gpnf-auto-attach-child-files.php
```diff
@@ -0,0 +1,71 @@
...
+
+	// If parent form has upload fields, rely on the notification's Attachments setting.
+	if( ! empty( $upload_fields ) ) {
+		if ( ! rgar( $notification, 'enableAttachments', false ) ) {
+			return $notification;
+		}
+	}
+	// Otherwise, rely on a manually defined array of notification IDs.
+	else {
+		$notification_ids = array( '5daaedb49dc32', '5dbce25cc21c2' );
```
spivurno, line 31: Let's add an inline instruction here to update these to the notification IDs to which this should apply. Notification IDs can be retrieved from the `nid` parameter when editing a notification. Confirm that's still true. 😅

gp-nested-forms/gpnf-auto-attach-child-files.php
```diff
@@ -0,0 +1,71 @@
+<?php
+/**
+ * Gravity Perks // Nested Forms // Auto-attach Uploaded Files from Child to Parent Notifications
+ * https://gravitywiz.com/documentation/gravity-forms-nested-forms/
+ *
+ * This snippet automatically attaches files uploaded onto the child form to the parent form notifications.
```
spivurno, line 6: Let's include some instructions here on how this works. If parent form has any File Upload fields, this snippet will rely on the "Attachments" setting for each parent notification. Otherwise, you must specify a list of notifications by ID to which child uploads should be attached.

spivurno (changes requested):

spivurno:
@sbassah Looks like @eihabi is working on a PR for this snippet as well. Could you add your instructional changes (and the plugin header) to his branch? 

https://github.com/gravitywiz/snippet-library/pull/168/files

PR #165, opened 2021-10-05
Auto-update Form ID on Form Duplication

sbassah:
Original Gist: https://gist.github.com/spivurno/91be22f6c6ca4ac4cfb104768b0694d2


sbassah:
I've added the plugin header.

PR #168, opened 2021-10-06
Fixed an issue where uploaded files were not attached to parent notification in `gp-nested-forms-auto-attach-child-files`.

eihabi:
This PR migrates the `gp-nested-forms-auto-attach-child-files` snippet to the library.

It also fixes a configuration issue where uploaded files were not attached to the parent form due to a notification ID check. By removing this check, once the snippet is enabled, any form that contains a child form will have uploaded files attached when this snippet is enabled.

Source: https://gist.github.com/spivurno/89efb788d3dc54a35ac001ad13322a74
Ticket: [#27868[(https://secure.helpscout.net/conversation/1647973886/27868?folderId=3808239)

### DIFF
--- a/gp-nested-forms/gp-nested-forms-auto-attach-child-files.php
+++ b/gp-nested-forms/gp-nested-forms-auto-attach-child-files.php
@@ -5,19 +5,18 @@
  */
 add_filter( 'gform_notification', function( $notification, $form, $entry ) {
 
-	if( ! class_exists( 'GPNF_Entry' ) ) {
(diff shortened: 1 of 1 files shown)


spivurno:
@eihabi Close! I was intending to rework the logic a bit. Here's a breakdown:

1. If specific notification IDs are provided, only check for those.
2. Otherwise, if parent form contains upload fields, check for parent notification's attachments settings.
3. If parent form does not contain upload fields, apply automatically.

eihabi:
@spivurno Done, new logic in the latest patch.

PR #205, opened 2021-11-10
Validate UK Postcode

sbassah:
As requested in this [ticket](https://secure.helpscout.net/conversation/1691274072/28775?folderId=3775888). Here's a simple version of the snippet to validate the postcode against the UK regex pattern.


spivurno (approved):
Ran this through PHPCBF and made a few minor tweaks. Otherwise, approved! ✅

PR #206, opened 2021-11-12
gpnf-hourly-cron.php

sbassah:
Original Gist: https://gist.github.com/spivurno/961aae8e182d5bee0b75c67807f596a2


spivurno (approved):

spivurno:
Just a note... I was going to request a plugin header for this one but I considered that most users probably will just be using this as an example of how to modify the cron timing rather than using this specific timing. So all good!

PR #216, opened 2021-11-23
Added `gpepee_edit_entry_id` filter to allow modifying the entry ID used to target the entry to be edited.

spivurno:
[HS#29155](https://secure.helpscout.net/conversation/1703886460/29155)

User wanted to conditionally edit the entry. This filter will allow him to do this with the following sample code:

```
add_filter( 'gpepee_edit_entry_id_123', function( $update_entry_id, $form_id ) {
	// Update "4" to the field ID you want to check the value of.
	if ( rgpost( 'input_1' ) === 'Add New' ) {
		$update_entry_id = false;
	}
	return $update_entry_id;
}, 10, 2 );
```

One sneaky but significant change I made in this PR (unrelated to the filter) is changing from which form the edit entry ID is fetched when fetching from the session. For some reason, it was setup to pull from the target form's session rather than the source form's session.

This doesn't make sense as the entire point of the snippet is to edit the same entry that has been populated into the form. I suspect most users are using the token or logged-in user's last entry option which were functioning correctly.

### DIFF
--- a/gp-easy-passthrough/gpep-edit-entry.php
+++ b/gp-easy-passthrough/gpep-edit-entry.php
@@ -21,10 +21,31 @@ class GPEP_Edit_Entry {
 		$this->form_id        = rgar( $options, 'form_id' );
 		$this->delete_partial = rgar( $options, 'delete_partial', true );
 
+		add_filter( "gpep_form_{$this->form_id}", array( $this, 'capture_passed_through_entry_ids' ), 10, 3 );
 		add_filter( "gform_entry_id_pre_save_lead_{$this->form_id}", array( $this, 'update_entry_id' ), 10, 2 );
 		add_filter( "gpls_rule_groups_{$this->form_id}", array( $this, 'bypass_limit_submissions' ), 10, 2 );
+
 		// Enable edit view in GP Inventory.
 		add_filter( "gpi_is_edit_view_{$this->form_id}", '__return_true' );
+
+	}
+
+	public function capture_passed_through_entry_ids( $form, $values, $passed_through_entries ) {
+
+		if ( empty( $passed_through_entries ) ) {
+			return $form;
+		}
+
+		// Add hidden input to capture entry IDs passed through via GPEP.
+
+		add_filter( "gform_form_tag_{$form['id']}", function( $form_tag, $form ) use ( $passed_through_entries ) {
+			$entry_ids = wp_list_pluck( $passed_through_entries, 'entry_id' );
+			$input     = sprintf( '<input type="hidden" name="%s" value="%s">', $this->get_passed_through_entries_input_name( $form['id'] ), implode( ',', $entry_ids ) );
+			$form_tag .= $input;
+			return $form_tag;
+		}, 10, 2 );
+
+		return $form;
 	}
 
 	public function update_entry_id( $entry_id, $form ) {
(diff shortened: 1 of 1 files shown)


spivurno:
@dariospace We worked together quite a bit on the previous version of this snippet. Would you mind reviewing my notes above and making sure everything still works for you?

@claygriffiths Would appreciate a quick code review from you.

@scyt Once this PR is merged, could we add a quick example to the docs about how to do this kind of conditional editing based on the example above?

dariospace:
@spivurno my guess is that it was fetching the data from the the target form's session because if not, the form will always show the last submitted entry data even if the checkbox to populate last submitted entry is unchecked.

[](https://www.loom.com/share/d553d20c914e4a8ead897e957d32d760)

When testing the filter, it always create a new entry, not sure if I'm using it correctly (shown in the video)

spivurno:
@dariospace That is correct. It will always populate the last entry which is saved in the GPEP cookie. If they enable the user's last submitted entry, that is specific to the WordPress user and only applies if the user is logged-in. This allows GPEP to work across different devices/sessions for logged-in users.

The functionality for retrieving the edit entry ID in this snippet should match GPEP's own functionality for fetching which entry is populated. I believe this change (from pulling from the target form's session to the source form's) aligns with this expectation.

As for the conditional snippet not working, you mentioned a video? Were you referring to mine or did you mean to include your own? 😄

spivurno:
Same form passthrough was previously disabled in GPEP. It's supported now so I've updated the snippet to handle it. 

This snippet has turned into a bit of nightmare because GPEP does not concisely handle the logic of which entry it will populate so this snippet has to do a ton of legwork to determine which entry GPEP is populating so it knows which entry to edit.

Hesitant to dump too much more time into this snippet since Entry Blocks will handle this so much better but I am going to follow up tomorrow with a list of known scenarios for how GPEP determines which entry to populate.

- [ ] ep_token
- [ ] ep_token - set but empty
- [ ] user's last entry
- [ ] default

spivurno:
Whew! This one got tricky and now it's super simple again. 😄

This PR now represents a major refactor (in addition to the new filter). Rather than trying to determine which entry GPEP used to populate the form, we let it tell us which entries were passed through and then capture those entry IDs in a hidden input that is preserved across validation errors and pagination.

All my testing was successful!

@dariospace Will you give this another pass to ensure the different types of entry passthrough are still working as expected? And then give the new filter another try as well?

dariospace:
@spivurno looks good 👍
The only thing I've found is that if a EP Token is used on a URL, even if it is removed afterwards, it seems the cookie is saved and the data from the form is repopulated but my guess is that this is default behavior.

1. Front End. Go to a form. See nothing is populated (no cookies are saved yet).
2. Front End. Go to the same form. Use an EP Token. Data is populated.
3. Front End. Go to the same form. Remove the EP Token from the URL. The data is still populated.

dariospace (commented):
Already commented on a comment

spivurno:
@dariospace Thanks for the review! And yes, this is how GPEP functions by default. This snippet will edit any entry populated by GPEP. If we have users who want to only edit if a token is present, it'll be simple enough to support. Let me know if it comes up!

gp-easy-passthrough/gpep-edit-entry.php
```diff
@@ -44,70 +65,39 @@ public function update_entry_id( $entry_id, $form ) {
...
-		}
-
-		return $rule_groups;
+	public function get_passed_through_entries_input_name( $form_id ) {
+		return "gpepee_passed_through_entries_{$form_id}";
 	}
 
-	public function get_update_entry_id( $target_form_id ) {
+	public function get_passed_through_entry_ids( $form_id ) {
+		$entry_ids = explode( ',', rgpost( $this->get_passed_through_entries_input_name( $form_id ) ) );
```
claygriffiths, line 73: I really dig how simple the code is but I do fear that this could allow overwriting any entry in the database (or at least for the same form).

spivurno: @claygriffiths Thoughts on my new approach? I've added a hash of the passed-through entry IDs which is posted through with the entry IDs themselves. When we fetch them we validate that the hash matches before returning the posted entry IDs.

PR #222, opened 2021-12-02
Require Unique Value Between Parent & Child

sbassah:
Original Gist: https://gist.github.com/spivurno/e7ea0631603b6bb7fa9a737cfd601f30


spivurno (approved):
Good work!

PR #226, opened 2021-12-21
added video walkthrough for GPPA All Results

dariospace:
I've created a video walkthrough for our spell crafted GPPA All Results snippet 👏

### DIFF
--- a/gp-populate-anything/gppa-populate-all-results.php
+++ b/gp-populate-anything/gppa-populate-all-results.php
@@ -6,6 +6,13 @@
  * By default, when populating a field's value, Populate Anything will only populated the first found result. This
  * snippet adds support for populating all found results. Just add the "gppa-populate-all" CSS class to the field's
  * CSS Class Name setting.
+ *
+ * Video
+ *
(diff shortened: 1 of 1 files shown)


spivurno (changes requested):
The length and general quality of the video is perfect for what we're trying to accomplish. I have a few suggestions for making the actual content more digestible.

1. Close all the extra tabs — it'll make it easier for the viewer to follow along.
2. Explain how you have your field setup — many users aren't going to be entirely familiar with Populate Anything so a very brief summary of how the field is configured to populate categories will go a long way.

    Bonus points if you show them the Results in the field settings so they can see what is available for population with the snippet.
    
3. Setup some simpler categories — it'll help the user understand what's happening and what's actually being populated into the field.

dariospace:
@spivurno I've created a new version per your comments and using Krisp to reduce audio background noise.

spivurno (approved):
Great stuff! 

You might consider hiding the Krispy UI in future videos but don't re-record this one. It's good to go. 🙂

scyt (approved):

PR #227, opened 2021-12-22
Added video walkthrough to populate custom DB with GPPA

dariospace:


scyt (changes requested):
This is great, but I think you should show how to populate from a separate, custom database. Currently you are populating from the WP database, and I think that can be confusing.

You can [build a new DB using adminer](https://kinsta.com/blog/adminer/#create-a-new-database), [add a new table](https://kinsta.com/blog/adminer/#create-tables-views-routines-and-events), and [insert records](https://kinsta.com/blog/adminer/#insert-new-records-and-update-existing-ones). 

Then, record a video showing the customer how to populate from that custom table.

dariospace:
I've added a custom DB, a table and dummy values to showcase how it could work.

scyt (approved):
LGTM!

PR #228, opened 2021-12-23
added video to showcase GW Validate Current Password snippet for GFUR

dariospace:
First time I've used this snippet! A cool one 🆒

### DIFF
--- a/gravity-forms/gw-validate-current-password.php
+++ b/gravity-forms/gw-validate-current-password.php
@@ -2,15 +2,10 @@
 /**
  * Gravity Wiz // Gravity Forms // Validate Current Password
  * https://gravitywiz.com/
+ * 
+ * Instruction Video: https://www.loom.com/share/118994cfff0f4441a519116bbc3ed0c7
  *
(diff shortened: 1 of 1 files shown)


scyt (approved):
I'm loving these. LGTM.

spivurno:
@dariospace I realized we don't have a standard way of including video links in our gist headers. Here's my hot take based on other examples in the Snippet Library:

https://github.com/gravitywiz/snippet-library/pull/228/commits/901cd988f107895de1f1bfc63df81194c4e6a6d8

I think the video is a high value resource so putting it higher in the description makes sense to me. It's easy enough to skip past if they just want to read the description. I think it also makes sense to give a clear, consistent label. I went with "Instruction Video" as it is both clear and the most common label already in use in the Snippet Library (though not by much).

Consider this a proposal. @scyt, welcome your thoughts here as well.

scyt:
@spivurno I dig it.

dariospace:
@spivurno perfect, I'll adopt that format.

PR #231, opened 2021-12-28
added Instruction Video for populating current quarter of the year

dariospace:
I've created this videos since I wasn't sure how to use the snippet. I've found an HS thread with the reference to use it [HS#21817](https://secure.helpscout.net/conversation/1386182183/21817/)

### DIFF
--- a/gravity-forms/gw-populate-current-quarter.php
+++ b/gravity-forms/gw-populate-current-quarter.php
@@ -2,6 +2,10 @@
 /**
  * Gravity Wiz // Gravity Forms // Populate the Current Quarter
  * https://gravitywiz.com/snippet-library/
+ *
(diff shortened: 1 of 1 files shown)


spivurno:
@dariospace Removed some extra lines from the snippet header here: https://github.com/gravitywiz/snippet-library/pull/231/commits/7ef702dd00db396db271302158cfa21d037f6607

Could you also capitalize "Video" in "Instruction video"?

dariospace:
Sure, done!

scyt (changes requested):
@dariospace Looks good. I think refreshing to reload the parameter can be a little confusing. It would probably be a good idea to keep the parameter in your clipboard so you can easily paste it in.

Could you redo this one without the refresh?

dariospace:
@scyt I've made a new video!

scyt (approved):
Thanks for making that change. Looks great. ✅

PR #237, opened 2022-01-05
added Instruction Video for Unique ID as Username

dariospace:


scyt (changes requested):
At 0:22 and 1:42 you say "Post Creation add-on". It should be "User Registration add-on".

This is approved with those changes. ✅

dariospace:
Added comments on the Loom video to warn the user about User Registration Add On

PR #243, opened 2022-01-10
Added support for selecting GPPA-populated choice fields in notifications.

spivurno:
[HS#29963](https://secure.helpscout.net/conversation/1728316556/29963/)

The previous logic here expected the value of the choices to be an email field in order to make it available for selection in the notification. This PR adds support for GPPA-populated fields.


claygriffiths (approved):
This is a nice pairing! 🍷🥩

PR #248, opened 2022-01-19
Get Form ID by Form Title

sbassah:
Original Gist: https://gist.github.com/spivurno/5cc99da16d877ba32616a368d1a60c88


spivurno (approved):
Loved your idea of the helper function with the more explicit title but in this case, I think using GF's function is straightforward enough. 🙂

PR #249, opened 2022-01-20
added Instruction Video gw-check-all-checkboxes-by-default

dariospace:


scyt (approved):
I like how you mentioned at the end that they can add multiple form IDs or just one. Anticipating a customer's question and answering it in the video is 💯.

Also, the audio quality of this video is really nice. Did you get a new microphone?

✅

dariospace:
Thanks for the kind words @scyt. I tunned a bit the setup for the audio, glad it´s being noticed 🎙

PR #255, opened 2022-01-27
Added gpi_remove_choices fitler

sbassah:
Although the exhausted choices were selectable, they were not submittable.  The additional gpi_remove_choices filter will ensure that the choices are not removed which will ensure that the validation is bypassed when the form is submitted.


spivurno (approved):
Approved for now! 

Clay is going to fix this in the near future so the remove choices filter won't be necessary since you can control whether choices are kept/removed via the field settings with GPI.

PR #258, opened 2022-02-01
added video for gpnf disable sessions

dariospace:


scyt (approved):
Looks good. At 1:30-1:40 you activate the snippet twice. Could you edit it one of those activations?

Approved with that change. ✅

PR #264, opened 2022-02-09
added Instruction Video convert time to dropdown

dariospace:


scyt (approved):
LGTM. Can you start recording these at the resolution we talked about and window only? We'd like to start mirroring these on youtube, and that should prevent the letterboxing effect.

@spivurno Should we be worried about the PHPCS failures on this?

PR #265, opened 2022-02-09
GW Time Sensitive Choices: Added new `remove_choices` setting to remove past choices instead of disabling them

claygriffiths:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/1777997481/31297?folderId=1530128

## Summary

Add a setting to remove past choices instead of disabling them.

### Screenshots

_Without choices removed_

CleanShot 2022-02-09 at 11 26 36

_Choices removed_

CleanShot 2022-02-09 at 11 27 09

### DIFF
--- a/gravity-forms/gw-time-sensitive-choices.php
+++ b/gravity-forms/gw-time-sensitive-choices.php
@@ -28,11 +28,12 @@ class GW_Time_Sensitive_Choices {
 
 		// set our default arguments, parse against the provided arguments, and store for use throughout the class
 		$this->_args = wp_parse_args( $args, array(
-			'form_id'        => false,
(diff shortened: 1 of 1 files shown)


spivurno (changes requested):
Works well! 

Only thing that feels a little funky is when there are no times available. The Drop Down is completely empty. [Quick video](https://gwiz.io/3B8VHd0). Can we add a default option in that case like, "No times available."?

spivurno (approved):
Looks good! 

Approved but if you're feeling frisky, providing an option to set the `&ndash; No times available &ndash;` text would be dank. I don't mind waiting until a customer actually requests it thought.

PR #271, opened 2022-03-01
added video append/prepend to merge tags

dariospace:


dariospace:
@scyt I've added a comment to clarify it's HTML!

PR #280, opened 2022-03-10
Dario/video/gw is in operator

dariospace:


spivurno (approved):
Haha, I love it! I think it'd be too much if we were posting this directly in the docs but these videos seem like a great place to let our silly side loose. 😄

spivurno:
Hilarious timing... the next email in my inbox was from Gusto with this GIF:

scyt (approved):

PR #283, opened 2022-03-15
added video for gpep delete entry after submission

dariospace:


scyt (commented):
Video is great, but the aspect ratio is off. Looks like you're recording the whole desktop on this one. Can you re-record at the pixel size we discussed and record only the browser window?

dariospace:
@scyt I updated the video to match the screen size required ✅

scyt (approved):

PR #286, opened 2022-03-17
Added checks to ensure that field exist before checking for input type

sbassah:
[Ticket](https://secure.helpscout.net/conversation/1818729044/32606?folderId=14964#thread-5355725313)

The snippet causes a fatal error if $field is null. This PR fixes that issue by checking if $field exists before it checks for the input type.


spivurno (changes requested):
As written, if there is no field, the code would continue to execute which looks like it would generate a similar error on line 37:

`$value = $field->get_value_merge_tag( ... );`

My instinct here would be to do something like this:

```
if ( ! $field || $field->get_input_type() !== 'date' ) {
    continue;
}
```

sbassah:
Your instinct is right. Your suggestion fixed the issue, Thanks.

PR #288, opened 2022-03-21
Require Minimum and Maximum Character Limit

sbassah:
Original Snippet: https://gist.github.com/spivurno/8220561#file-gw-gravity-forms-minimum-characters-php

Per the [request of a user](https://secure.helpscout.net/conversation/1820941677/32691?folderId=3775888#thread-5362256382), I added support to use the snippet on multi-input fields


spivurno (approved):
Good stuff!

PR #289, opened 2022-03-24
~ added video for gpaa require selection

dariospace:


scyt (changes requested):
Video is approved, but all of the single quotes were replaced with double quotes in this commit. Can you revert those back to single quotes?

PR #303, opened 2022-04-19
~ added instruction video

dariospace:


scyt (approved):
✅ 

The audio on this one is a little overdriven. I think it's acceptable, but you might want to turn down the gain/input level on your mic for future videos.

PR #309, opened 2022-05-03
Dynamically Populating User Role

dariospace:
Original Gist: https://gist.github.com/spivurno/3710924


scyt (approved):
Approved with changes:

There's some whitespacing issues on lines 16 and 17. There's a space before the tab. The spaces should be removed.

PR #312, opened 2022-05-04
Skip Pages on Multi-Page Forms

dariospace:
Original Gist: https://gist.githubusercontent.com/spivurno/3710898

PR #315, opened 2022-05-05
~ replaced space with tabs l8-10

dariospace:
Replaced spaces with tab per comment: https://github.com/gravitywiz/snippet-library/pull/312


scyt (approved):
Looks like they were still spaces. Code editors can be stubborn. I replaced these with tabs in the github UI.

PR #322, opened 2022-05-09
Improved support for validating field groups via AJAX.

spivurno:
[HS#34494](https://secure.helpscout.net/conversation/1872966406/34494/)

The `field_map` parameter allows you to pass an array of fields that should all match the same entry from the source form. This worked correctly on submission but the AJAX validation validated each field individually which incorrectly implied that an entry with both values had been found.

This PR introduces some dramatic changes to add support for ensuring that the AJAX validation works the same way as the on-submission validation.


spivurno:
@dariospace Could you give this a look and let me know how this works for you based on this ticket?

https://secure.helpscout.net/conversation/1872966406/34494/

spivurno:
Closed in favor of https://github.com/gravitywiz/snippet-library/pull/346.

PR #326, opened 2022-05-11
Added instruction video

dariospace:


scyt (changes requested):
At 0:32, you should also update the `value`. Otherwise, Gravity Forms is going to store `static` in the form entry. I suggest updating it to match the label, so both are `Add New`.

scyt (approved):

PR #329, opened 2022-05-19
Added instruction Video

dariospace:


scyt (changes requested):
Looks good, except for the small big from :48-1:02 where the form is unresponsive. I think you can just cut that out of the video and it should still work.

PR #332, opened 2022-05-27
added Instruction Video

dariospace:


scyt (approved):
At 0:12 you say "populating", but I think you mean "selecting". Is it easy to replace that? 

If not, go ahead and merge as-is.

PR #334, opened 2022-05-28
Added better support for handling pages with no fields after randomization.

spivurno:
[HS#34770](https://secure.helpscout.net/conversation/1882336037/34770)

Customer was originally seeking a way to show the "Submit" button when the last page with fields had been reached. Previously, this snippet handled empty pages by automatically adjusting the target page to skip them on submission. It was feasible to dynamically change the name of the "Next" button when we knew that subsequent pages had no fields; however, this would not solve the issue where the pagination still showed additional pages.

I considered using JavaScript to interact with the pagination conditional logic to update the page navigation and button; however, I was honestly kind of curious about simply restructuring the form pages during the randomization rather than having to interact with the pagination conditional logic at all.

The basic logic is...

1. Filter the fields based on the specified options.
2. Get all the page numbers for the remaining fields (e.g. 1, 3, 4, 6).
3. Loop through the page numbers and reassign them to their new page numbers based on the index of the page numbers array.
4. As part of that loop, we also Inject the page field into the filtered fields and modify their page numbers accordingly.

I'm fairly happy with the result but I do have some concerns about brittleness. This has been tested with the customer's form extensively and it works with pages with only random fields, only static fields, and both random and static fields on the same page. Excited to see how this works in the wild!

**Edit: Sept 30th, 2023**

This ticket is requesting the same functionality (hiding pages with no fields) but in the context of Entry Blocks. Whatever lessons we find here can likely be applied there.

https://secure.helpscout.net/conversation/2373971553/55284/


DamianSmith85:
Having issues with multipage form using this updated script.

The final page break field is sometimes called in upon refresh and I receive duplicate "Next" buttons.
The form cannot be completed due to this. It's a 50/50 whether it gets called in and breaks the form.

See below, should only be 8 questions being called in, but the random page break at the start sets it as 9.

The page break ID (46) is not included in the array:
`new GFRandomFields( 4, 8, array( 3, 9, 7, 5, 18, 19, 20, 22, 29, 30, 31, 32, 33, 34, 35, 36 ), true );`

Screenshot 2023-05-26 at 13 50 37

spivurno:
@DamianSmith85 Drop us a line [via support](https://gravitywiz.com/support/) and we'll be happy to dig in. 🙂

PR #342, opened 2022-06-09
GPPS Live Refresh

sbassah:
Original Snippet: https://gist.github.com/spivurno/4bbfcd02ac99b89087f1e7b417ef1a5f


spivurno (approved):
Great work, @sbassah! 

I made one tweak to remove the "GP" from the perk name. We do this since we're already indicating it's part of the "Gravity Perks" suite.

https://github.com/gravitywiz/snippet-library/pull/342/commits/592e8cad69f1aee25164b0119f1b72bf12083d09

PR #351, opened 2022-07-13
Set a Custom Validation CSS Class

sbassah:
Hook Documentation: https://gravitywiz.com/documentation/gppt_script_args/#set-a-custom-validation-css-class


spivurno (changes requested):
I don't understand what this snippet original purpose was or how it would actually work. Let's scrap it for now and remove it from the documentation page.

sbassah:
Noted.

PR #352, opened 2022-07-13
Set a Custom Transition Style

sbassah:
Hook Documentation: https://gravitywiz.com/documentation/gppt_script_args/#set-a-custom-transition-style


spivurno (changes requested):
This can be scrapped. `scrollUp` is no longer supported.

sbassah:
Sure.

PR #353, opened 2022-07-13
Create gppa_lmt_whitelist.php

dariospace:
Added hook to the library


spivurno (approved):
✅

Made two tweaks:

1. Updated file name to use dashes instead of underscores.
2. Converted to use the anonymous function format.

Make note of the first but don't worry about converting future snippets to the anonymous function format. One of the devs can handle that. 🤘

PR #355, opened 2022-07-13
Adding a Custom Transition Style

sbassah:
Hook Documentation: https://gravitywiz.com/documentation/gppt_transition_styles/#adding-a-custom-transition-style


spivurno (changes requested):
We've updated the library since we first implemented this and "Scroll Up" is no longer supported. Let's scrap this and remove it from the documentation.

sbassah:
Noted.

PR #363, opened 2022-07-14
Create gpls-field-validation-message.php

dariospace:
added hook


spivurno (approved):
✅

1. Removed the description.
2. Add update instructions contextually above the filter name.
3. Updated the form ID and field ID to our standard "123" and "4".
4. Updated title to be sentence case (first half was, second half wasn't).

PR #364, opened 2022-07-14
Create gpwc-word-count

dariospace:
added hook


spivurno (approved):
✅

Made a couple tweaks:

1. Update title to describe what this specific snippet does.
2. Ditched the description since it was now equivalent to the title.
3. Replaced spaces with tabs.

PR #365, opened 2022-07-14
Create gpui-unique-id

dariospace:
added hook


spivurno (changes requested):
Let's scrap this snippet and remove the example from the docs. I believe this predates the addition of the "Prefix" setting in the Unique ID field settings.

PR #367, opened 2022-07-15
Added Support to set WC Meta value with field value.

sbassah:
https://secure.helpscout.net/conversation/1949115202/36695?folderId=3775888


spivurno (approved):
Beautiful!

spivurno:
Went ahead and merged this so I could send it on over to the customer. Thank you, @sbassah!

PR #370, opened 2022-07-15
Change Default Form Page

sbassah:
Hook Documentation: https://gravitywiz.com/documentation/gpmpn_default_page/#set-default-page-to-second-page


spivurno (approved):
✅

Moved the description inline as an instruction for how the snippet can be configured.

PR #377, opened 2022-07-18
Create gpnf-session-initialized.js

dariospace:
added hook


spivurno (changes requested):
Let's scrap this one as we have a superior method for accomplishing this here: https://github.com/gravitywiz/snippet-library/blob/master/gp-nested-forms/gpnf-dyn-pop-session-hash.php

PR #378, opened 2022-07-18
Create gpml-supported-field-types.php

ghost:
added hook


spivurno (approved):
✅

1. Updated title to sentence case and removed mention of "GP Media Library" since the snippet already contextualizes the perk previously in the title.
2. Fixed miscellaneous PHPCS issues.

PR #379, opened 2022-07-19
Disable Passing Through Entries Without a Token

sbassah:
Hook Documentation: https://gravitywiz.com/documentation/gpep_bypass_session_init/#disable-passing-through-entries-without-a-token


spivurno (approved):
✅

PR #385, opened 2022-07-19
Create gpnf-fetch-form-html-after-add.js

dariospace:
added hook


spivurno (approved):
✅

Since we're instructing the user to use our Custom Javascript plugin, I've removed the `<script>` blocks which would cause an error.

PR #387, opened 2022-07-19
Create gpnf-jquery-ui-dependencies.php

dariospace:
added hook


spivurno (changes requested):
Let's scrap this one since jQuery UI is fully deprecated now.

PR #388, opened 2022-07-19
Create gpcp-after-update-pricing.js

dariospace:
added hook


spivurno (approved):
✅

Typically we title-case the title but given how long this one is, I think you made the write call to leave it as sentence-case. 👍

PR #393, opened 2022-07-19
Scale Uploaded Images

sbassah:
Hook Documentation: https://gravitywiz.com/documentation/gpfup_image_loader_options/#scale-instead-of-cropping-based-on-dimensions-set-in-the-cropping-settings


spivurno (approved):
✅

PR #399, opened 2022-07-20
 Include Image Metadata/EXIF on Uploaded images

sbassah:
Hook Documentation: https://gravitywiz.com/documentation/gpfup_strip_image_metadata/#include-image-metadata-exif-on-all-uploaded-images


spivurno (approved):
✅

PR #402, opened 2022-07-20
Add Number as a Supported Field Type

sbassah:
Hook Documentation:
https://gravitywiz.com/documentation/gpi_supported_field_types/#add-number-as-a-supported-field-type


spivurno (approved):
✅

PR #403, opened 2022-07-21
gp-pay-per-word fixes counting bug

JanStorm:
The string splitting functions cannot handle non-ascii bytes like umlauts (äöü) or emojis and would generate more than one array element for them.
In result, this whole function would return a character count of 6 (or 5 in JS) for this String: `Hiö🍻`, which is only 4 long.

The solution would return 4 as length of the string above in both cases.

See this link for a Demo on the PHP-site: https://www.tehplayground.com/PdKHf9u7UNoYH6c5
And the JSFiddle for the js-site: https://jsfiddle.net/xjrqczh4/


spivurno:
@JanStorm Epic! Thank you. 😃

PR #409, opened 2022-07-21
Override Limit of Specific Choice

sbassah:
Hook Documentation: https://gravitywiz.com/documentation/gplc_choice_limit/#filter-a-specific-choice-8217-s-limit


spivurno (approved):
✅

Updated single quotes in comment to double quotes.

PR #414, opened 2022-07-22
Removed experimental snippet that has since been published

scyt:
[experimental/gpnf-populate-parent-form-id-in-child-form.js](https://github.com/gravitywiz/snippet-library/blob/master/experimental/gpnf-populate-parent-form-id-in-child-form.js) has been migrated to [gp-nested-forms/gpnf-populate-parent-entry-id.js](https://github.com/gravitywiz/snippet-library/blob/master/gp-nested-forms/gpnf-populate-parent-entry-id.js).

This PR removes the experimental version.


spivurno (commented):
✅

spivurno (approved):

PR #417, opened 2022-07-22
Create gppr-price-range-min-max.php

dariospace:
added hook


spivurno (approved):
✅

Some more feedback here: https://www.loom.com/share/506fb45781ba49c8bdcecb8b70b0e75d

PR #422, opened 2022-07-26
Create gppa-increase-query-limit.php

dariospace:
added hook


dariospace:
It gives this error gp-populate-anything/gppa-increase-query-limit.php#L7

Tabs must be used to indent lines; spaces are not allowed

But on Line 7 there is a Tab already

spivurno (approved):
✅

spivurno:
@dariospace Here's an explanation of the PHPCS warning you were seeing. 🙂

https://www.loom.com/share/170500b57a41404dae06043b2bfa7d12

PR #454, opened 2022-08-01
 Send Notification on the last day of every month

sbassah:
Hook Documentation: https://gravitywiz.com/documentation/gpns_schedule_timestamp/#send-notification-on-the-last-day-of-every-month


spivurno (commented):
✅

1. Let's use "1a23bc456def7" as our standard NID sample value.
2. Updated "Form Id" to "form ID".
3. Update "Notification ID" to "notification ID".

spivurno (approved):

PR #456, opened 2022-08-01
Change Scope and Unsubscribe Success URL

sbassah:
Hook Documentation: https://gravitywiz.com/documentation/gpns_unsubscribe_url_args/#change-scope-and-unsubscribe-success-url


spivurno (commented):
✅

1. Updated FORMID and NID to example values in the filter name. Mixed feelings on this for the NID as it's such a weird value but my thinking is that it's helpful for them to understand the type of value they're looking for.

spivurno (approved):

PR #457, opened 2022-08-01
Update gpwc-disable-truncation.php

dariospace:
added instruction video


scyt (changes requested):
This one is good, but the volume is pretty low. Is there any way to increase the volume in Loom without having to re-record?

dariospace:
@scyt I've added a new video to the snippet with improved sound

scyt (changes requested):
I hate to make you record this one again, but the aspect ratio is off. Can you redo this one with only the window being captured?

scyt (approved):

PR #458, opened 2022-08-01
gw-user-registration-update-by-email.php: Added new capability setting.

claygriffiths:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/1959544702/37054?folderId=1530128

## Summary

Provide more control over capabilities for Update by Email. Prior to this PR, it required a user to be logged in with no way to change the behavior.

This PR sets a new required capability of `edit_users` for anyone accessing these forms. This is more strict than the previous requirement of any logged-in user being able to edit.


spivurno (approved):
Clever. I like it.

PR #472, opened 2022-08-02
Create gpnf-add-confirmation-to-delete-in-parent.js

dariospace:
added hook


spivurno (approved):
✅

2. Tweaked title, filename, and description for brevity + clarity.

PR #473, opened 2022-08-02
Create gw-all-fields-exclude-fields-with-personal-info.php

dariospace:
added hook


dariospace:
@spivurno I'm not sure why this is failing, I've modify several times but it keeps showing errors

PR #475, opened 2022-08-03
Added styling support for displaying images as a grid.

spivurno:
Also, removed support for displaying non-image file types and made links mandatory.

By limiting the scope of this we can simplifying the styling requirement significantly and still meet the requesting user's requirements. Until I share this and he wants it displayed the exact same way as GPFUP. 😅

Thoughts?


claygriffiths (approved):
I dig it!

Honestly, I think we can drop the `wp_kses` call here. That was for Gravity PDF.

PR #477, opened 2022-08-09
`gw-first-error-focus`: Fixed improper scrolling issue on multipage forms

veryspry:
# Context

https://secure.helpscout.net/conversation/1963934628

# Summary

- Added `.editorconfig`
- Fixed issue with multi-page forms where validation error that is no longer active would be scroled to.
- Fixed issue where multi-page forms would not scroll to the top.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,13 @@
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+indent_style = tab
+insert_final_newline = true
+trim_trailing_whitespace = true
+indent_size = 4
+
+[*.yml]
+indent_style = space
+indent_size = 2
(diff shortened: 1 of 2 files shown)


claygriffiths (approved):

spivurno:
@veryspry First PR! Woot, woot! 

Could you add a quick inline comment above the new code that scrolls to the top of the form if no error is present? I didn't understand the necessity for this without reading the ticket and it'll be helpful for future us. 😄

claygriffiths:
> Could you add a quick inline comment above the new code that scrolls to the top of the form if no error is present? I didn't understand the necessity for this without reading the ticket and it'll be helpful for future us. 😄

Here we are 17 months later 🤣

---

The original request:

> The scroll to error works great, however, the script disables the anchoring of the form to the top when there is no error present. We are using a multi-page ajax form and when the user goes to the next page, they are essentially at the bottom of the form.

Dario's escalation

image

veryspry:
> > Could you add a quick inline comment above the new code that scrolls to the top of the form if no error is present? I didn't understand the necessity for this without reading the ticket and it'll be helpful for future us. 😄

> Here we are 17 months later 🤣

I really failed that one l 😆

PR #481, opened 2022-08-11
Fixed issue in `gp-nested-forms/gpnf-triggered-population` where nested form entries were not removed when unselecting the configured checkbox

veryspry:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/1973533007/37541?folderId=6987275

## Summary

This snippet is supposed to add child form entries when a configured checkbox / radio button is selected and then remove that child entry item when de-selected.

There are a couple of items at play here that this PR addresses:

1. This was originally written to support radio buttons only and didn't work correctly with multi-select. The reason is because we weren't checking to see if the field was `checked` before calling `addChildEntry` in the `onchange` handler. Thus, when the field was de-selected (`!checked`), the add function would still be called.
2. The `removeChildEntry` function had not been implemented yet. Thus, even when it was called it didn't actually do anything.

Please note that this now relies on some new changes within the `gp-nested-forms` perk and consequently will require a new version of that perk to work correctly.

### DIFF
--- a/gp-nested-forms/gpnf-triggered-population.php
+++ b/gp-nested-forms/gpnf-triggered-population.php
@@ -81,7 +81,7 @@ class GPNF_Triggered_Population {
 							var value = input.val();
 							var checked = input[0].checked;
 
-							if ( checked && value === self.triggerFieldValue || ( value !== '' && self.triggerFieldValue === '_notempty_' ) ) {
+							if ( (checked && value === self.triggerFieldValue) || ( value !== '' && self.triggerFieldValue === '_notempty_' ) ) {
 								self.addChildEntry();
 							} else {
 								self.removeChildEntry();
@@ -164,7 +164,6 @@ class GPNF_Triggered_Population {
 
 		return empty( $this->_args['form_id'] ) || $form_id == $this->_args['form_id'];
 	}
-
 
 	public function ajax_add_child_entry() {
 
--- a/gp-nested-forms/gpnf-triggered-population.php
+++ b/gp-nested-forms/gpnf-triggered-population.php
@@ -77,8 +77,11 @@ class GPNF_Triggered_Population {
 						$form = $( '#gform_{0}'.format( self.formId ) );
 
 						$( '#field_{0}_{1}'.format( self.formId, self.triggerFieldId ) ).find( 'input' ).on( 'change', function() {
-							var value = $( this ).val();
-							if ( value === self.triggerFieldValue || ( value !== '' && self.triggerFieldValue === '_notempty_' ) ) {
+							var input = $( this );
+							var value = input.val();
+							var checked = input[0].checked;
+
+							if ( checked && value === self.triggerFieldValue || ( value !== '' && self.triggerFieldValue === '_notempty_' ) ) {
 								self.addChildEntry();
 							} else {
 								self.removeChildEntry();
@@ -98,6 +101,8 @@ class GPNF_Triggered_Population {
 
 						$.post( self.ajaxUrl, request, function( response ) {
 							if ( response.success ) {
+								// store the entry data for later for usage in the removeChildEntry method
+								window.gpnf_triggered_population_entry = response.data
 								GPNestedForms.loadEntry( response.data );
 							}
 						} );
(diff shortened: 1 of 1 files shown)


gp-nested-forms/gpnf-triggered-population.php
```diff
@@ -98,14 +101,25 @@ public function output_script() {
...
 								GPNestedForms.loadEntry( response.data );
 							}
 						} );
 
 					}
 
 					self.removeChildEntry = function() {
-						// @todo
+
+						var entryDataRowElem = $("tr[data-entryid='" + window.gpnf_triggered_population_entry.entryId + "']");
```
veryspry, line 114: Using oldschool string interop here for ES5 support

gp-nested-forms/gpnf-triggered-population.php
```diff
@@ -215,18 +230,14 @@ public function get_instance_hash() {
...
 }
 
-# Configuration
-
 new GPNF_Triggered_Population( array(
-	'form_id'              => 123,
-	'trigger_field_id'     => 4,
-	'trigger_field_value'  => 'Yes',
-	'nested_form_field_id' => 5,
+	'form_id'              => 8,
```
veryspry, line 234: Will revert these values!

gp-nested-forms/gpnf-triggered-population.php
```diff
@@ -215,8 +230,6 @@ public function get_instance_hash() {
 
 }
 
-# Configuration
-
```
claygriffiths, line 219: Can you re-add this?

claygriffiths (approved):
Awesome work, and great write-up!

Once the customer confirms this is working, I recommend a Squash & Merge to clean up the commit message.

claygriffiths:
Oh, you'll also want to increment the `Version:` at the top of the file prior to merging.

gp-nested-forms/gpnf-triggered-population.php
veryspry: :100:

gp-nested-forms/gpnf-triggered-population.php
```diff
@@ -77,8 +77,11 @@ public function output_script() {
 						$form = $( '#gform_{0}'.format( self.formId ) );
 
 						$( '#field_{0}_{1}'.format( self.formId, self.triggerFieldId ) ).find( 'input' ).on( 'change', function() {
-							var value = $( this ).val();
-							if ( value === self.triggerFieldValue || ( value !== '' && self.triggerFieldValue === '_notempty_' ) ) {
+							var input = $( this );
+							var value = input.val();
+							var checked = input[0].checked;
+
+							if ( checked && value === self.triggerFieldValue || ( value !== '' && self.triggerFieldValue === '_notempty_' ) ) {
```
claygriffiths, line 84: Does this `checked && value === self.triggerFieldValue` need parentheses?

claygriffiths (approved):
Nice work! Just one potential change I saw.

gp-nested-forms/gpnf-triggered-population.php
```diff
@@ -150,6 +165,7 @@ public function is_applicable_form( $form ) {
 		return empty( $this->_args['form_id'] ) || $form_id == $this->_args['form_id'];
 	}
 
+
```
claygriffiths, line 168: Ideally, whitespace changes (additions/removals) aren't snuck in unless it's fixing a code style. I won't hold up the PR over this! Just something to be aware of.

gp-nested-forms/gpnf-triggered-population.php
veryspry: Ooh, yeah probably

gp-nested-forms/gpnf-triggered-population.php
veryspry: Got it! Will remove :+1:

gp-nested-forms/gpnf-triggered-population.php
veryspry: Added in because I _think_ it makes it a bit more robust and still works correctly with the added parens

PR #485, opened 2022-08-15
Include ALL HTML Fields in All Fields Merge Tags

sbassah:
Original Snippet: https://secure.helpscout.net/conversation/1306745488/20339/#thread-3771604868


spivurno:
@sbassah Did a new customer need this specific snippet? If so, do you think using the All Fields Template plugin would be a workable solution for them?

https://gravitywiz.com/gravity-forms-all-fields-template/#field-filtering-modifier

The only catch is that they would have to specifically include the IDs of each HTML field.

sbassah:
@spivurno Yes, someone requested it in the [comment section](https://gravitywiz.com/gravity-forms-all-fields-template/comment-page-5/#comment-829589). According to them, they have over 80 forms with HTML fields and they wanted the HTML fields to show without having to enter include the IDs of each HTML field in all the forms.  Reading through their question again, I am not sure if this snippet would even work for them.

spivurno (approved):
✅

Tweaked the snippet to not require the modifier. Install this and it'll include HTML fields in the {all_fields} merge tag output by default.

sbassah:
Thanks, David. I will update the comment reply.

sbassah:
@spivurno I just tested the snippet with the {all_fields} merge tags. It doesn't work unless I change this ` $tag !== 'all_fields' ` to this `$tag == 'all_fields' `.

spivurno:
Ah, my bad. Go ahead and make the change and merge this bad boy. 😄

PR #487, opened 2022-08-16
Update gpwc-disable-limits-for-logged-in-users.php

dariospace:
## Context

added instruction video


scyt (changes requested):
Can you redo this one with only the window being captured? The aspect ratio is off and we don't need to see the menubar and dock.

dariospace:
@scyt new video added with the correct screen ratio, no dock, no menu bar

scyt (approved):

PR #489, opened 2022-08-16
Update gpfup-disable-submit-during-upload.js

dariospace:
## Context

added instruction video


scyt (changes requested):
Video is great. Can you record it again with only the window captured?

dariospace:
@scyt new video added with the correct screen setup.

scyt (approved):

PR #490, opened 2022-08-16
`gpi-waiting-list.php`: Refactor to class and add support for Single Products

claygriffiths:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/1972333442/37514/

## Summary

* Refactor to class
* Add support for waiting list on Single Products (previously only choice-based)

### DIFF
--- a/gp-inventory/gpi-waiting-list.php
+++ b/gp-inventory/gpi-waiting-list.php
@@ -8,98 +8,192 @@
  * Replace the default available inventory message with a waiting list message when a choice's inventory is exhausted.
  * The choice will be selectable and submittable.
  */
-add_filter( 'gpi_disable_choices', '__return_false' );
-add_filter( 'gpi_remove_choices', '__return_false' );
+class GPI_Waiting_List {
+	public $waitlist_message = '(waiting list)';
 
-function gpiw_apply_waitlist_message( $choice, $field, $form, $how_many_left = false ) {
-	$message         = '(waiting list)';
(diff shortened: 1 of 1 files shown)


gp-inventory/gpi-waiting-list.php
```diff
@@ -8,99 +8,197 @@
...
+		 */
+		add_filter( 'gpi_disable_choices', '__return_false' );
+		add_filter( 'gpi_remove_choices', '__return_false' );
+
+		add_filter( 'gpi_pre_render_choice', array( $this, 'pre_render_choice' ), 10, 5 );
+
+		// Add support for showing waiting list message in confirmations and notifications.
+		add_filter( 'gform_pre_submission_filter', array(
+			$this,
+			'add_waitlist_message_to_choices_on_submission'
```
spivurno, line 32: I'm assuming this got written like this due to a line-length rule? When the queue is less busy, could we revisit the topic of excluding this rule? I find it so disruptive to visually parsing the code when this happens. And it hurts my feelings. 😭

gp-inventory/gpi-waiting-list.php
```diff
@@ -8,99 +8,197 @@
...
+		), 10, 2 );
+
+		add_filter( 'gform_entries_field_value', function ( $value, $form_id, $field_id, $entry ) {
```
spivurno, line 35: Not a deal breaker but it does feel a little strange to be using anonymous functions now that this is a class?

spivurno (approved):
Awesome, man! I probably should have converted this to a class on my last passthrough.

Approved with some comments inline.

gp-inventory/gpi-waiting-list.php
claygriffiths: I think this was PhpStorm reformatting. Fixing

PR #491, opened 2022-08-18
Added snippet that modifies the nested form button text content

veryspry:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/1976741186/37641/

## Summary

This adds a snippet that allows you to modify the "add" button message next to nested form entry tables.

As an example: 
image

### DIFF
--- a/gp-nested-forms/gpnf-nested-entries-max-message.php
+++ b/gp-nested-forms/gpnf-nested-entries-max-message.php
@@ -1,0 +1,51 @@
+<?php
+/**
+ * Gravity Perks // Nested Forms // Add Child Entry on Trigger
+ * https://gravitywiz.com/documentation/gravity-forms-nested-form/
+ *
+ * Instruction Video: https://www.loom.com/share/2d01000744354e7693ac4348f521992f
+ *
+ * This snippet allows overrides the default button message in the Nested Form Perk
+ * and shows the user the minimum and maximum number of entries that can be added.
+ *
+ * Plugin Name:  GP Nested Forms - Show Minimum and Maximum Entries Message
+ * Plugin URI:   https://gravitywiz.com/documentation/gravity-forms-nested-form/
+ * Description:  Override the message next to child field submit buttons and show the minimum/maximum number of child entries allowed.
+ * Author:       Gravity Wiz
+ * Version:      0.2
+ * Author URI:   https://gravitywiz.com
+ */
+
+
+function modify_child_form_button_messsage( $args ) {
+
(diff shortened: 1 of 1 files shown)


spivurno (changes requested):
Dude. This is a very exciting PR. Gives me a lot of insight into the kind of developer you are and I like it! 

1. Love the thought you put into making this as reusable as possible. Some tweaks on this below.
2. Love the initiative in recording an instruction video. These aren't required but are always welcomed. 🙏

We'll want to make some changes here for consistency with our other snippets.

1. Most of our customers are very familiar with adding hooks directly so it'll likely be easier for them to bind their own hooks. Rather than having the `override_child_form_max_entry_message()` helper method for registering the hooks, let's let the customer register their own hooks to the `template_args_filter` function.
2. Currently, we're adding 3 new functions to the global scope. We'd preferably only add one or none at all by using an anonymous function. Example:

    ```php
	// Update "123" to your form ID and "4" to your field ID. Remove "_123_4" to apply this globally.
	add_filter( 'gpnf_template_args_123_4', function( $args, $form_field, $form ) {
 
		// $args->add_button_message is not always present when this hook is applied
		if ( ! array_key_exists( 'add_button_message', $args ) ) {
			return $args;
		}
 
		$min = $form_field->gpnfEntryLimitMin;
		$max = $form_field->gpnfEntryLimitMax;
	        $message = '';
 
		if ( ( empty( $min ) || $min === '0' ) && ! empty( $max ) ) {
			$message =  'You can add no more than ' . $max . ' entries.';
		} elseif ( ( empty( $max ) || $max === '0' ) && ! empty( $min ) ) {
			$message = 'You must add at least ' . $min . ' entries.';
		} elseif ( ! empty( $min ) && ! empty( $max ) ) {
			$message = 'You must add at least ' . $min . ' entries and no more than ' . $max . ' entries.';
		}

        	$message = sprintf(
			'
		 	<p class="gpnf-add-entry-max">
		 		%s
	 		</p>',
			$message
		);
 
		return $args;
	}, 10, 3 );

3. In the example above, you'll also notice that I merged the `format_message` function's functionality into the now anonymous `template_args_filter` function. This removes another function from the global scope and keeps the snippet simpler for the end user who often doesn't need to know the exact mechanics of how it all works so long as they can easily understand how to customize it.
4. Oh, just noticed this has a plugin header. Since it's a smaller snippet and requires configuration, we typically wouldn't include the plugin header as this will likely be installed via a snippet manager or directly in their theme's functions.php to more easily allow them to modify it for there needs.

claygriffiths (approved):

PR #493, opened 2022-08-19
Fixed `gw-choice-counter` bug where count would not update when choices were populated from GP Populate Anything.

veryspry:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/1975203407/37592?folderId=6987275

## Summary

The choice counter snippet was not working when counting from a field populated by Populate Anything. This was because, as the field changed (was repopulated), we did not update / rebind the event handler functions so when the choice fields changed (e.g. were checked) the count wouldn't get updated.

### DIFF
--- a/gravity-forms/gw-choice-counter.php
+++ b/gravity-forms/gw-choice-counter.php
@@ -63,13 +63,7 @@ class GW_Choice_Count {
 						}
 					}
 
-					self.init = function() {
-
-						// Event handler for all listeners to avoid DRY
-						var updateChoiceEventHandler = function() {
-							self.updateChoiceCount( self.formId, self.choiceFieldIds, self.countFieldId );
-						};
-
+					self.updateEventHandlers = function() {
 						for( var i = 0; i < self.choiceFieldIds.length; i++ ) {
 
 							var choiceFieldId       = self.choiceFieldIds[i],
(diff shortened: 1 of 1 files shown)


gravity-forms/gw-choice-counter.php
```diff
@@ -63,34 +63,42 @@ function output_script() {
...
-
+					self.updateEventHandlers = function() {
 						for( var i = 0; i < self.choiceFieldIds.length; i++ ) {
 
 							var choiceFieldId       = self.choiceFieldIds[i],
 								choiceFieldSelector = '#input_' + self.formId + '_' + choiceFieldId,
 								$choiceField        = $(choiceFieldSelector),
 								$parentForm         = $choiceField.parents('form');
 
+							$parentForm.off( 'click', choiceFieldSelector, self.updateChoiceEventHander );
```
veryspry, line 74: Clean up existing event handlers

claygriffiths: I dig it!

One pro-tip here if getting a reference to the callback is impossible (or a lot of work) is to leverage jQuery event namespaces.

e.g. `$parentForm.on( 'click.gwChoiceCounter', choiceFieldSelector, self.updateChoiceEventHandler );`

then you can call...

e.g. `$parentForm.off( '.gwChoiceCounter' );`

claygriffiths (approved):
Solid!

Feedback on the title: perks should be `GP [Perk Name]` in most references. GW is an old naming format. You may also see `Gravity Forms [Perk Name]` or `GF [Perk Name]`, but this is reserved for the website.

gravity-forms/gw-choice-counter.php
veryspry: Smooooth :sunglasses:

PR #498, opened 2022-08-24
Update gppa-explode-commas-as-choices.php

dariospace:
## Context
added instruction video


scyt (changes requested):
This is a solid video, but the browser window wasn't maximized and I can see your email inbox at the bottom.

Can you record this one again with the browser window maximized?

dariospace:
@scyt thanks! I just made a new video without emails and the correct screen size 😊

scyt (approved):

PR #500, opened 2022-08-29
Updated the snippet to also set the value property

sbassah:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/1992124058/38187?folderId=14964

The snippet wasn't setting the value property of the choice-based field when populated. When a multi-select field is used, it displays the value property of the field on the entry page, and because the value property hasn't been set by the snippet it displays the entry ID. Added a line of code to also set the value.


spivurno (approved):
Good work, @sbassah!

spivurno (approved):

PR #507, opened 2022-09-02
Bump version from 0.5 -> 0.6 in gw-save-and-continue-auto-load

veryspry:
## Context

I forgot to bump the version number in a snippet :facepalm:


claygriffiths:
Follow-up to https://github.com/gravitywiz/snippet-library/pull/501

PR #508, opened 2022-09-02
`gpnf-override-parent-merge-tag-on-submission.php`: reprocess calculations on submission based on changed parent form values

claygriffiths:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/1991241515/38165?folderId=3808239

## Summary

Reprocess calculated values in a Nested Form when using the `gpnf-override-parent-merge-tag-on-submission.php` snippet. The calculated fields can be excluded using `$exclude_field_ids`.


spivurno (approved):
Dayum. Sneaky, sneaky.

This looks good. It does make me wonder if we need to account for changes in the order data when Pricing fields are involved but I say let's wait for a customer to report that scenario.

PR #509, opened 2022-09-05
`gw-gravity-forms-rename-uploaded-files.php`: Fixed issue where forward slashes and dots were not being removed from user inputted filenames.

veryspry:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/1994083346/38231?folderId=6987275

## Summary

We were not removing "/" or "." chars from new file names inputted by users in the `gw-gravity-forms-rename-uploaded-files.php` snippet.

The result is that users could ultimately manipulate the directory which a file was uploaded to which is insecure.

### DIFF
--- a/gravity-forms/gw-gravity-forms-rename-uploaded-files.php
+++ b/gravity-forms/gw-gravity-forms-rename-uploaded-files.php
@@ -207,7 +207,7 @@ class GW_Rename_Uploaded_Files {
 
 	function rename_file( $file, $entry ) {
 
-		$new_file = $this->get_template_value( $this->_args['template'], $file, $entry );
+		$new_file = $this->get_renamed_filepath( $this->_args['template'], $file, $entry );
 		$new_file = $this->increment_file( $new_file );
 
 		return $new_file;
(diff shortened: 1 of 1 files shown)


gravity-forms/gw-gravity-forms-rename-uploaded-files.php
```diff
@@ -241,32 +241,28 @@ function is_path( $filename ) {
...
-		$value = str_replace( '{filename}', $info['filename'], $template );
+		$filename = str_replace( '{filename}', $info['filename'], $template );
 
 		// replace merge tags
-		$form  = GFAPI::get_form( $entry['form_id'] );
-		$value = GFCommon::replace_variables( $value, $form, $entry, false, true, false, 'text' );
+		$form     = GFAPI::get_form( $entry['form_id'] );
+		$filename = GFCommon::replace_variables( $template, $form, $entry, false, true, false, 'text' );
+		// make sure filename is "clean". This includes removing any user inputted items such as "../", "/usr/bin" etc
+		$filename = sanitize_file_name( $filename );
```
veryspry, line 254: It was really annoying me that `$this->clean()` was just aliasing another function :laughing:

spivurno: Haha, I feel ya. The reason this it was pulled out into its own function is to allow folks to extend the class and provide their own cleaning function depending on their needs. We'll need to preserve it for that reason and backwards compat.

veryspry: Got it! Will revert.

I am curious, given that this is a snippet and folks are manually copying this into their site somewhere, why would someone extend the class to override the `clean` method rather than just modifying the code where the string cleaning happens?

It seems like an odd pattern, and I'd love to understand the benefits.

@spivurno

gravity-forms/gw-gravity-forms-rename-uploaded-files.php
```diff
@@ -241,32 +241,28 @@ function is_path( $filename ) {
...
-
 		// replace our custom "{filename}" psuedo-merge-tag
-		$value = str_replace( '{filename}', $info['filename'], $template );
+		$filename = str_replace( '{filename}', $info['filename'], $template );
 
 		// replace merge tags
-		$form  = GFAPI::get_form( $entry['form_id'] );
-		$value = GFCommon::replace_variables( $value, $form, $entry, false, true, false, 'text' );
+		$form     = GFAPI::get_form( $entry['form_id'] );
+		$filename = GFCommon::replace_variables( $template, $form, $entry, false, true, false, 'text' );
```
veryspry, line 252: The biggest change here is that the merge tags are now replaced before concatenating the user-inputted filename with the rest of the file path.

This allows us to clean/sanitize the filename completely before it's added to the file path whereas adding it to the file path first, before cleaning, makes it impossible to know if all user input has been cleaned.

gravity-forms/gw-gravity-forms-rename-uploaded-files.php
spivurno: Good question! 

Since we continue to make updates/fixes to the snippet, altering the class directly makes it more difficult for them to copy-and-paste the latest version. By extending the class and leaving the original class intact, they can safely update it in the future while preserving their own changes.

spivurno:
Leaving the technical review to @claygriffiths but I did want to mention that whatever changes we make here will need to be duplicated in the upcoming perk. Here's a good card to keep track of this: https://www.notion.so/gravitywiz/GPFR-1-0-Final-Push-45fcd50d90934a8d806bb96e6317b2a5

veryspry:
We're just waiting on the customer to confirm this works for them before merging! They are on holiday now and will test on their return

veryspry:
The customer confirmed this is working correctly for them. @claygriffiths can I get a review? :smile:

veryspry:
Hmmm, looks like `phpcs` is complaining about an unrelated file :thinking:

veryspry:
The user confirmed that this works correctly :+1: 

@claygriffiths or @saifsultanc I'd love a review when one of you has time!

claygriffiths (approved):

PR #510, opened 2022-09-07
Fixed issue where child entry changes did not trigger a GPPA refresh when traversing AJAX-enabled forms.

spivurno:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/1995729085/38292/

## Summary

Customer was experiencing an issue populating child entry data into the parent form when traversing a multi-page AJAX form with conditional logic enabled. I was able to reproduce the same issue when an AJAX-enabled form was reloaded with a validation error.

The root issue is that the "gpnf_session_initialized" action that this snippet relies on is not triggered for subsequent loads of an AJAX-enabled form. I've updated the code to check if the session has already been initialized and if so to setup the Parent Entry ID field to trigger change events per usual.


claygriffiths (approved):
Nice work!

PR #511, opened 2022-09-08
Update gpnf-add-child-entry-total-in-entry-list.php

dariospace:
## Context
added instruction video


scyt (changes requested):
Video is good, but needs to be redone with only the window captured.

dariospace:
@scyt new video added, it should match the guidelines.

scyt (approved):

PR #521, opened 2022-09-13
`gppa-populate-checkboxes-as-choices.php`: Added Instruction Video and Updated Snippet Header.

levisimon:
## Summary
`gppa-populate-checkboxes-as-choices.php`: Added Instruction Video and Updated Snippet Header.

### DIFF
--- a/gp-populate-anything/gppa-populate-checkboxes-as-choices.php
+++ b/gp-populate-anything/gppa-populate-checkboxes-as-choices.php
@@ -1,21 +1,27 @@
 <?php
 /**
- * See https://gravitywiz.com/documentation/how-do-i-install-a-snippet/ for details on how to install snippets like these.
+ * Gravity Perks // Populate Anything // Use Checkbox Field as Choices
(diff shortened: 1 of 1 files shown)


scyt (changes requested):
I think the best demo for this is to populate into a Radio Buttons or a Drop Down field. Populating into a checkbox works, but in my experience most users are populating into another choice type.

I'm also curious about changing FORMID and FIELDID to 123 and 4. Tagging @spivurno for his input there.

spivurno:
Using "123" and "4" for form ID and field ID is our current practice. All good there!

spivurno (approved):

dariospace:
@scyt should we approve this one? I'm not sure what would be the next step in the Workflow

scyt:
@dariospace Could you record a new version where you populate into a Radio Buttons field? That better matches the use case of this snippet.

Also, make sure to use the new template for the commit message. You'll also need to edit the PR name.

PR #524, opened 2022-09-19
Added support for using "weekday" and "weekdays" keyword in modifier.

spivurno:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2010411216/38704?folderId=14965

## Summary

Customer desired to modify a user-entered date by a given number of weekdays and display that date immediately. The Populate Dates snippet supports modifying dates "live"; however, it previously did not support the "weekday" keyword.

Example configuration:

```
new GW_Populate_Date( array(
	'form_id'         => 123,
	'source_field_id' => 4,
	'target_field_id' => 5,
	'modifier'        => '+7 weekdays',
) );
```

## Todos

- [ ] Update article.

### DIFF
--- a/gravity-forms/gw-populate-date.php
+++ b/gravity-forms/gw-populate-date.php
@@ -1115,7 +1115,7 @@ class GW_Populate_Date {
 											break;
 										case 'weekday':
 										case 'weekdays':
-											// todo
+											this.rwd += amount;
 											break;
 									}
 								}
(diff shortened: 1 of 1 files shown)


gravity-forms/gw-populate-date.php
```diff
@@ -1418,6 +1420,18 @@ function processTzCorrection(tzOffset, oldValue) {
 								result.setFullYear(this.y, this.m, this.d);
 								result.setHours(this.h, this.i, this.s, this.f);
 
+								if ( this.rwd ) {
+									var mod = this.rwd < 0 ? -1 : 1;
```
claygriffiths, line 1424: What does the `r` in this stand for?

Also, what is this line doing specifically?

claygriffiths (commented):
Good stuff! Takes reading through a couple of times to figure out what's going on, but I think that may be the nature of the DateTime beast.

claygriffiths (approved):

gravity-forms/gw-populate-date.php
spivurno: Haha, I have no idea. Just followed suit with the naming pattern of all the other variables of this type.

I added a comment to provide some clarity. Essentially, we need to add or subtract a day for each weekday allowance specified in the `this.rwd` variable. We know it should subtract if `this.rwd` is a negative number. Below, this `mod` variable is used to automatically handle whether a day is added or subtracted.

claygriffiths: Nice, thank ya!

PR #530, opened 2022-09-22
Added snippet to Default Check First Radio Button having Randomized Choices

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2010508229/38708/#thread-5970679827

## Summary

Functionality for Radio Button populated with GPPA and having 'Randomize Choice Order' (GP Randomizer) to have the first choice as the default selection.

### DIFF
--- a/gp-populate-anything/gppa-select-first-random-choice-on-radio.js
+++ b/gp-populate-anything/gppa-select-first-random-choice-on-radio.js
@@ -1,0 +1,11 @@
+/**
+ * Gravity Perks // Populate Anything // Default the First Populated Choice for Radio having choices populated with GPPA
+ * https://gravitywiz.com/documentation/gravity-forms-populate-anything/
+ *
+ * 1. Install this snippet with our free Custom JavaScript plugin.
+ *    https://gravitywiz.com/gravity-forms-custom-javascript/
+ */
+
+// Update field ID to match your Radio Button.
+var fieldID = 4; 
+$('input:radio[name=input_' + fieldID + ']')[0].checked = true;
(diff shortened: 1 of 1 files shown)


spivurno (changes requested):
One thing I just thought of is the need to make sure we're targeting this input in a way that is form specific. Thoughts on using the `id` attribute instead of the name and then using the `GFFORMID` "constant" provided by our Custom Javascript plugin?

Another consideration is simplifying the actual code by allowing the field ID to be replaced directly in the snippet rather than creating a field ID. If this were a more complex snippet I think a separate variable would add clarity so don't consider this a hard rule. 😄

Untested Example:

```js
// Update "4" to the the ID of your Radio Button field.
$( '#input_GFFORMID_4' )[0].checked = true;
```

saifsultanc:
> One thing I just thought of is the need to make sure we're targeting this input in a way that is form specific. Thoughts on using the `id` attribute instead of the name and then using the `GFFORMID` "constant" provided by our Custom Javascript plugin?
> 
> Another consideration is simplifying the actual code by allowing the field ID to be replaced directly in the snippet rather than creating a field ID. If this were a more complex snippet I think a separate variable would add clarity so don't consider this a hard rule. 😄
> 
> Untested Example:
> 
> ```js
> // Update "4" to the the ID of your Radio Button field.
> $( '#input_GFFORMID_4' )[0].checked = true;
> ```

@spivurno From what I understand, with the `id` attribute, we will have to use _attribute starts with selector_, because each radio button ID is concatenated with a `$choice_id` counter (L248 on class-gf-field-radio). I just tested it, and this should be it:

```
// Update "4" to the the ID of your Radio Button field.
$( 'input[id^=choice_' + GFFORMID + '_4_' )[0].checked = true;
```

Lmk, if I should push this update?

spivurno:
Since the randomization is happening on the query level, the first choice rendered would always be `choice_FORMID_FIELDID_0` but doing it with the attribute starts selector would make this compatible with GP Randomizer out of the box, so let's roll with it!

spivurno (approved):

PR #531, opened 2022-09-23
Disable selected dropdown choice in other dropdown

sbassah:
I had this old JS snippet in my Gist, so I wanted to move it to the experimental folder to make it easily accessible. Its functionality is to disable selected dropdown choices in other dropdown fields that have the same set of choices.


spivurno (approved):
Awesome, man!

I applied some formatting and renamed. ✅

PR #532, opened 2022-09-23
Updated "Require Unique Values for Different Fields" snippet with case sensitive option

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2015924398/38889?folderId=3808239

## Summary

[Gravity Forms: Require Unique Values for Different Fields](https://gravitywiz.com/gravity-forms-require-unique-values-for-different-fields/#download-snippet) makes the case-insensitive comparison. This update adds the case-sensitive argument, with its default as false.

### DIFF
--- a/gravity-forms/gw-require-unique-values.php
+++ b/gravity-forms/gw-require-unique-values.php
@@ -132,9 +132,7 @@ class GW_Require_Unique_Values {
 	public function get_value_hash( $value ) {
 
 		if ( ! $this->_args['case_sensitive'] ) {
-			foreach( $value as &$v ) {
-				$v = strtolower( $v );
-			}
+			$value = array_map( 'strtolower', $value );
 		}
 
 		// Replace values like "1.1" with "x.1" to make it generic for comparison.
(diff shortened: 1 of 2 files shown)


gravity-forms/gw-require-unique-values.php
```diff
@@ -130,6 +131,12 @@ public function get_filtered_value( $field ) {
 
 	public function get_value_hash( $value ) {
 
+		if ( ! $this->_args['case_sensitive'] ) {
+			foreach( $value as &$v ) {
+				$v = strtolower( $v );
+			}
```
claygriffiths, line 137: FWIW, this can be simplified using `array_map`:

```suggestion
			$value = array_map( 'strtolower', $value );
```

claygriffiths (approved):
Nice work! Just one small comment, you can take it or leave it. 

The `foreach` allows for more conditionals in the future, like ensuring `$v` is a string prior to running `strtolower` on it.

PR #533, opened 2022-09-27
`gpnf-override-parent-merge-tag-on-submission`: Updated to a class and made configuration for multiple child forms easier.

veryspry:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2015154047/38861/ 

It is difficult to configure `gpnf-override-parent-merge-tag-on-submission.php` for multiple child forms. This enables users to more easily apply the logic to multiple child forms.

## Summary

- Refactors the snippet to a class to match our preferred snippet style.
- Support instantiating one instance of `GPNF_Override_Parent_Merge_Tags` per child form a user wants to configure.
- Support instantiating a single instance of `GPNF_Override_Parent_Merge_Tags` with no arguments to apply this to all child forms.

### DIFF
--- a/gp-nested-forms/gpnf-override-parent-merge-tag-on-submission.php
+++ b/gp-nested-forms/gpnf-override-parent-merge-tag-on-submission.php
@@ -7,66 +7,150 @@
  *
  * Override all {Parent} merge tags when the parent form is submitted or a parent entry is updated.
  */
-function gpnf_override_parent_merge_tags( $entry, $form ) {
-	// Update '123' to the ID of the Child form. Set as false to apply to all child forms.
-	$id_of_child_form = 123;
-	// Updated '4, 5, 6' with the IDs of the fields you want to skip. Leave blank to apply to all fields.
-	$exclude_field_ids = array( 4, 5, 6 );
 
-	foreach ( $form['fields'] as $field ) {
-		if ( $field->get_input_type() === 'form' ) {
-			$child_form_id = $field->gpnfForm;
-			if ( $id_of_child_form ) {
-				if ( $child_form_id != $id_of_child_form ) {
-					return $entry;
-				}
-			}
-			$child_form = GFAPI::get_form( $child_form_id );
-			foreach ( $child_form['fields'] as $child_field ) {
-				if ( in_array( $child_field->id, $exclude_field_ids ) ) {
-					continue;
-				}
-				if ( $child_field->get_entry_inputs() ) {
-					foreach ( $child_field->get_entry_inputs() as $input ) {
-						gpnf_override_child_entry_input_value( $entry, $field, $input['id'], rgar( $input, 'defaultValue' ) );
-					}
-				} else {
-					gpnf_override_child_entry_input_value( $entry, $field, $child_field->id, $child_field->defaultValue );
-				}
-			}
+class GPNF_Override_Parent_Merge_Tags {
+
+	/**
+	 * @var int|null child form ID to apply the override to. If null, override will apply to all child forms.
+	 */
+	private $child_form_id = null;
+
+	/**
+	 * @var int[] array field ID's to exclude when overriding a child form's {Parent} merge tags.
+	 */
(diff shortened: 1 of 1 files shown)


gp-nested-forms/gpnf-override-parent-merge-tag-on-submission.php
```diff
@@ -7,66 +7,115 @@
...
+	}
+
+	function set_config_data( $config ) {
+		if ( empty( $config ) ) {
+			return;
+		}
+
+		$this->child_form_id = $config['child_form_id'];
+
+		if ( array_key_exists( 'exclude_field_ids', $config ) ) {
```
veryspry, line 35: Unfortunately `rgar` might not be found by the interpreter at this point since this could be run before gravity forms is.

veryspry:
I added an additional config option called `include_field_ids`. In order to keep it simple, `exclude_field_ids` will be ignored if `include_field_ids` is used.


@spivurno Let me know what you think of this "API" change!

spivurno:
@veryspry Love it!

gp-nested-forms/gpnf-override-parent-merge-tag-on-submission.php
claygriffiths: I dig! I think in another PR I mentioned that we typically go with `isset()` or `! empty()` for brevity. I'm not gonna hold the PR up over this though.

gp-nested-forms/gpnf-override-parent-merge-tag-on-submission.php
```diff
@@ -7,66 +7,156 @@
...
+
+	/**
+	 * @var int[] array field ID's to exclude when overriding a child form's {Parent} merge tags.
+	 */
+	private $exclude_field_ids = array();
+
+	/**
+	 * @var int[] array field ID's to include when overriding a child form's {Parent} merge tags. If included, will override any usage of $exclude_field_ids.
+	 */
+	private $include_field_ids = array();
```
claygriffiths, line 26: Some spicy DocBlocks here! Thanks for adding these.

gp-nested-forms/gpnf-override-parent-merge-tag-on-submission.php
```diff
@@ -7,66 +7,156 @@
...
+# exclude_field_ids  (array) optional array of field IDs to exclude from the override.
+# included_field_ids (array) optional array of field IDs to apply this to. If included, then exclude_field_ids will be ignored.
+# -------------------------------------------------
+
+# Example - apply to all fields of a specific child form
+// new GPNF_Override_Parent_Merge_Tags(
+// 	array(
+// 		'child_form_id' => 1,
+// 	)
+// );
```
claygriffiths, line 143: Can you put the `array` on the same line as the class constructor call per feedback in that other PR? Someday we'll have a lint or Dangerfile rule for this 😅

claygriffiths (approved):
Nice work! Just one small tweak for usage comments, otherwise LGTM

gp-nested-forms/gpnf-override-parent-merge-tag-on-submission.php
veryspry: Yup, I believe so. I wrote this the same day (before your other review) and didn't even think about updating here so thanks for the reminder!

veryspry: Yup, I believe so. I wrote this the same day (before your other review) and didn't even think about updating here so thanks for the reminder!

gp-nested-forms/gpnf-override-parent-merge-tag-on-submission.php
veryspry: Done!

A dangerfile rule would be sick :fire:

veryspry:
We are currently waiting for a user to confirm that it works for them. After that, I'll merge!

PR #535, opened 2022-09-28
`gpnf-auto-attach-child-files`: Refactored for greater configuration flexibility.

veryspry:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2011825626/38759/

## Summary

- Refactors `gpnf-auto-attach-child-files` into a class.
- Moves configuration logic to constructor args so that users can easily create multiple, unique configurations.
- Adds `child_form_upload_field_ids` config option to limit which child form upload fields are included in a notification.

### DIFF
--- a/gp-nested-forms/gpnf-auto-attach-child-files.php
+++ b/gp-nested-forms/gpnf-auto-attach-child-files.php
@@ -11,68 +11,167 @@
  * Plugin URI:   http://gravitywiz.com/documentation/gravity-forms-nested-forms/
  * Description:  Auto-attach Uploaded Files from Child to Parent Notifications
  * Author:       Gravity Wiz
- * Version:      0.1
+ * Version:      0.2
  * Author URI:   https://gravitywiz.com/
  */
-add_filter( 'gform_notification', function( $notification, $form, $entry ) {
-	// Notes:
-	// - If notifications IDs are specified this snippet will only work for those specific forms.
-	// - If the parent form has upload fields, they will be attached instead of the child form's fields.
-	// - If neither of these conditions are present, child form uploads will be attached by default.
-	// Configuration:
-	// 1) Ensure that the parent form's notification has the "Attach uploaded fields to notification" checked.
-	// 2) [Optional] Modify the following array to limit the snippet to specific notification IDs.
-	// Notification IDs can be retrieved from the nid parameter in the URL when editing a notification.
-	// Example: $notification_ids = array( '5daaedb49dc32', '5dbce25cc21c2' );
-	$notification_ids = array();
+class GPNF_Auto_Attach_Child_Files {
+	/**
+	 * The ID of the parent form to apply this to (optional).
+	 */
+	private $parent_form_id = null;
 
-	if ( ! class_exists( 'GPNF_Entry' ) ) {
-		return $notification;
+	/**
+	 * The IDs of the child form field to apply this to (optional).
+	 */
+	private $child_form_field_ids = null;
+
+	/**
+	 * The IDs of the parent form notifications to which child uploads should be attached (optional).
+	 */
+	private $notification_ids = array();
+
+	/**
+	 * The IDs of child form fields whose files should be attached to notifications (optional).
+	 */
+	private $child_form_upload_field_ids = array();
+
+	function __construct( $config = array() ) {
+		$this->set_config_data( $config );
+		$this->init_hooks();
 	}
 
-	$upload_fields = GFCommon::get_fields_by_type( $form, array( 'fileupload' ) );
+	function init_hooks() {
+		add_filter( 'gform_notification', array( $this, 'handle_notification_attachments' ), 10, 3 );
+	}
 
-	if ( ! empty( $notification_ids ) && ! in_array( $notification['id'], $notification_ids ) ) {
-		return $notification;
-	} else {
+	function set_config_data( $config ) {
+		if ( ! empty( $config['parent_form_id'] ) ) {
+			$this->parent_form_id = $config['parent_form_id'];
+		}
+
+		if ( ! empty( $config['child_form_field_ids'] ) ) {
+			$this->child_form_field_ids = $config['child_form_field_ids'];
+		}
+
+		if ( ! empty( $config['notification_ids'] ) ) {
+			$this->notification_ids = $config['notification_ids'];
+		}
+
+		if ( ! empty( $config['child_form_upload_field_ids'] ) ) {
+			$this->child_form_upload_field_ids = $config['child_form_upload_field_ids'];
+		}
+	}
+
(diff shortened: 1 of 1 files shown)


gp-nested-forms/gpnf-auto-attach-child-files.php
```diff
@@ -11,68 +11,134 @@
...
+
+# Attach all child form uploads to all notifications.
+// new GPNF_Auto_Attach_Child_Files();
+
+# Attach all uploads from child forms to *only* the notification with id "63331f3fcf7f0".
+// new GPNF_Auto_Attach_Child_Files(
+// 	array(
+// 		'notification_ids' => array( '63331f3fcf7f0' ),
+// 	)
+// );
```
claygriffiths, line 129: Just for the sake of consistency with our other snippets, can you put the `array` on the same line as the class name?

e.g.

```suggestion
# Attach all uploads from child forms to *only* the notification with id "63331f3fcf7f0".
// new GPNF_Auto_Attach_Child_Files( array(
//    'notification_ids' => array( '63331f3fcf7f0' ),
// ) );
```

gp-nested-forms/gpnf-auto-attach-child-files.php
```diff
@@ -11,68 +11,134 @@
...
-	if ( ! empty( $notification_ids ) && ! in_array( $notification['id'], $notification_ids ) ) {
-		return $notification;
-	} else {
-		// If parent form has upload fields, rely on the notification's Attachments setting.
-		if ( ! empty( $upload_fields ) ) {
-			if ( ! rgar( $notification, 'enableAttachments', false ) ) {
-				return $notification;
-			}
+	function set_config_data( $config ) {
+		if ( array_key_exists( 'notification_ids', $config ) ) {
```
claygriffiths, line 32: FWIW, `array_key_exists( 'notification_ids', $config )` can be safely (no PHP notices) shorted to one of the following:

```php
if ( isset( $config['notification_ids'] ) ) {
```

or

```php
if ( ! empty( $config['notification_ids'] ) ) {
```

or even shorter (no if statement)...

```php
$this->notification_ids = rgar( $config, 'notification_ids', array() );
```

claygriffiths (commented):
Awesome work!

I'm going to give this a functional test here shortly. In the meantime, here are a couple of minor suggestions for the code.

gp-nested-forms/gpnf-auto-attach-child-files.php
veryspry: Noice!

Unfortunately `rgar` isn't always present since this seems to be instantiated before Gravity Forms but I think one of `isset` or `empty` will work :+1:

gp-nested-forms/gpnf-auto-attach-child-files.php
veryspry: Yup! Done

gp-nested-forms/gpnf-auto-attach-child-files.php
veryspry: Done

claygriffiths: Good call there! If you ever do need GF functions (or functions from perks), you can hook a method to the `plugins_loaded` action.

gp-nested-forms/gpnf-auto-attach-child-files.php
```diff
@@ -11,68 +11,128 @@
...
+						$attachments[]  = $attachment_url;
+					}
 				}
 			}
 		}
+
+		return $notification;
+	}
+
+	function is_applicabale_notification( $notification ) {
```
claygriffiths, line 93: Typo in this method name 😅

claygriffiths (approved):
Just noticed a typo in a method name and now PHPCS is barking about a couple of things.

I did a functional test and it looks good! I did notice that it slowed down submission a bit, but I'm not sure if that's an issue with the refactor itself.

I went ahead and approved this pending the two small things mentioned above.

gp-nested-forms/gpnf-auto-attach-child-files.php
veryspry: :stuck_out_tongue:

veryspry:
> I did a functional test and it looks good! I did notice that it slowed down submission a bit, but I'm not sure if that's an issue with the refactor itself.

Thanks for doing that! All was looking good during my tests but it's always nice to get an extra set of eyes on it :raised_hands: 

As far as the slowdown goes, my guess is that it isn't due to the refactor as the logic is essentially the same as before. The main difference is the way that configuration options are set

If I had to guess why, database reads are probably the main culprit. For instance:

- getting the form object which queries the form table and the form meta table: https://github.com/gravitywiz/snippet-library/pull/535/files#diff-6dc898ef07f19e6f8f6a1c641d2c979e25e78c5327ea4622eccf6b3eb35fb4c7R65
- getting all entries for each child form: https://github.com/gravitywiz/snippet-library/pull/535/files#diff-6dc898ef07f19e6f8f6a1c641d2c979e25e78c5327ea4622eccf6b3eb35fb4c7R66

We also make a call to `preg_replace()` for each attachment we find which might impact speed noticeably?

That said, PCRE is supposed to cache compiled regex's which should be helpful since we pass the same regex string to each call of `preg_replace()`

https://stackoverflow.com/a/210086

veryspry:
Just waiting on confirmation from a user that this works well for them before merging!

gp-nested-forms/gpnf-auto-attach-child-files.php
veryspry: Fixed

veryspry:
Feature request: 

Support for multiple child forms (in one parent form) whose upload fields have different id's.

For example: Parent Form has two child forms, `A` and `B`. Both `A`'s file upload field has id of `1` and `B`'s has a file upload field with id of `2`.

I'm going to go ahead and add that functionality to this snippet before merging!

veryspry:
Feature request: 

Support for multiple child forms (in one parent form) whose upload fields have different id's.

For example: Parent Form has two child forms, `A` and `B`. Both `A`'s file upload field has id of `1` and `B`'s has a file upload field with id of `2`.

I'm going to go ahead and add that functionality to this snippet before merging!

veryspry:
@claygriffiths just a heads up that I changed the API here a bit so that this works with more complex child form configurations.

See here for more details: https://github.com/gravitywiz/snippet-library/pull/535#issuecomment-1265688832

claygriffiths (approved):

veryspry:
We're just waiting for a user to confirm this works

veryspry:
This is confirmed to work as intended. I'm mergin' it!

PR #536, opened 2022-09-28
`gravity-forms/gw-limit-multiselect.js`: Fixed issue where multi select limit did not work correctly when going backwards on multi-page forms.

veryspry:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2020716657/39063/

## Summary

The issue was occuring because the `limitMultiSelect()` logic was only running when the configured select field changed. This meant that when returning to the select on a multi-page form, the limit logic may not have run yet and thus the user would be able to select more than the limit.

The fix is to run `limitMultiSelect()` each time the configured select comes into view.

### DIFF
--- a/gravity-forms/gw-limit-multiselect.js
+++ b/gravity-forms/gw-limit-multiselect.js
@@ -4,26 +4,35 @@
  *
  * Limit how many options may be selected in a Multi Select field. Works with
  * regular Multi Select fields as well as fields with Enhanced UI enabled.
- * 
+ *
  * Instructions:
- * 
+ *
  * 1. Install this snippet with our free Custom JavaScript plugin.
  *    https://gravitywiz.com/gravity-forms-custom-javascript/
  * 2. Configure snippet per inline instructions.
  */
-// Update "1" to the ID of your Multi Select field.
-$( '#input_GFFORMID_1' ).on( 'change', function () {
-	// Update "2" to the max number of options that should be selectable.
+(function() {
+	// Update "1" to the ID of your Multi Select field.
+	var multiSelectId = '#input_GFFORMID_1';
 	var maxSelected = 2;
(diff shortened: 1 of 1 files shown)


gravity-forms/gw-limit-multiselect.js
```diff
@@ -4,26 +4,35 @@
...
- * 
+ *
  * 1. Install this snippet with our free Custom JavaScript plugin.
  *    https://gravitywiz.com/gravity-forms-custom-javascript/
  * 2. Configure snippet per inline instructions.
  */
-// Update "1" to the ID of your Multi Select field.
-$( '#input_GFFORMID_1' ).on( 'change', function () {
-	// Update "2" to the max number of options that should be selectable.
+(function() {
```
veryspry, line 14: Popping this into an IIFE for variable scoping

claygriffiths: @veryspry I dig the thinking! However, since we recommend installing this with our Custom JS plugin, the scope should already be limited: https://github.com/gravitywiz/snippet-library/blob/matt%2Ffix-multi-select-limit/gravity-forms/gw-gravity-forms-custom-js.php#L176-L176

veryspry: I should have noticed since `$ === window.jQuery` in this context :laughing:

veryspry: Fixed!

veryspry:
This change fixes a legitmate bug, but does not help the user who this bug report initially came from as they are using frontend form library that changes multi-select dom structure and is incompatible with this snippet.

@claygriffiths Are you okay if I merge?

claygriffiths (approved):
@veryspry Yup! Thanks for adding this.

PR #539, opened 2022-10-05
`gpmpn-render-all-page-links-on-load.js`: Added new snippet to load all multi page navigation links when editing entries with GPEP.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2025453715/39250?folderId=3808239

## Summary

With GPEP + GPMPN if we use the [snippet](https://github.com/gravitywiz/snippet-library/blob/master/gp-easy-passthrough/gpep-edit-entry.php) for GPEP Edit entry, the form entry load appears as if it still creating a new entry (not editing). This snippet reloads all the links, so the edit entry look is there.


claygriffiths:
@saifsultanc For a commit message/PR title that matches the guidelines, you could use something like the following when you add a new snippet:

```
`gpmpn-render-all-page-links-on-load.js`: Added new snippet to load all multi page navigation links when editing entries with GPEP
```

claygriffiths (approved):
LGTM! Just one note about PR title/commit message.

PR #540, opened 2022-10-07
`gw-limit-checkboxes.php`: Migrated from GitHub Gists.

scyt:
## Context

📓 Notion: https://www.notion.so/gravitywiz/Limiting-How-Many-Checkboxes-Can-Be-Checked-e7b0bdd3c3094ceeb2ae941b7f1228a1

## Summary

Migrating [from gist](https://gist.github.com/spivurno/4005684#file-gistfile1-php).


scyt:
This one is going to throw some PHPCS errors. I tried to resolve them, but in doing so the snippet stopped working.

PR #544, opened 2022-10-12
`gpnf-child-products-in-parent-order-summary.php`: Updated snippet to process Live Merge Tags for Nested Form Fields.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2022459557/39108?folderId=3808239

## Summary

Nested Form Fields with live merge tags are not rendered completely. This fix includes the live merge tag processing for those fields.

### DIFF
--- a/gp-nested-forms/gpnf-child-products-in-parent-order-summary.php
+++ b/gp-nested-forms/gpnf-child-products-in-parent-order-summary.php
@@ -31,7 +31,7 @@ add_filter( 'gform_product_info', function( $product_info, $form, $entry ) {
 				continue;
 			}
 
-			$child_form    = gp_nested_forms()->get_nested_form( $field->gpnfForm );
+			$child_form    = gp_nested_forms()->get_nested_form( $nested_form_field->gpnfForm );
 			$_entry        = new GPNF_Entry( $entry );
 			$child_entries = $_entry->get_child_entries( $nested_form_field_id );
 
@@ -40,6 +40,16 @@ add_filter( 'gform_product_info', function( $product_info, $form, $entry ) {
 				$_child_products    = array();
 				foreach ( $child_product_info['products'] as $child_field_id => $child_product ) {
 					$child_product['name'] = "{$product_info['products'][ $field->id ]['name']} — {$child_product['name']}";
+
+					// If Nested Form fields have Live Merge Tags, process those.
+					if ( method_exists( 'GP_Populate_Anything_Live_Merge_Tags', 'has_live_merge_tag' ) ) {
+						$gppa_lmt = GP_Populate_Anything_Live_Merge_Tags::get_instance();
+						if ( $gppa_lmt->has_live_merge_tag( $child_product['name'] ) ) {
+							$gppa_lmt->populate_lmt_whitelist( $child_form );
+							$child_product['name'] = $gppa_lmt->replace_live_merge_tags_static( $child_product['name'], $child_form, $child_entry );
+						}
+					}
+
 					$_child_products[ "{$nested_form_field_id}.{$child_entry['id']}.{$child_field_id}" ] = $child_product;
 				}
 				$child_products = $child_products + $_child_products;


gp-nested-forms/gpnf-child-products-in-parent-order-summary.php
```diff
@@ -40,6 +40,19 @@
...
+						$gp_lmt         = GP_Populate_Anything_Live_Merge_Tags::get_instance();
+						$nested_form_id = $field->formId;
+						$nested_form    = GFAPI::get_form( $nested_form_id );
+						$matches        = array();
+						preg_match( $gp_lmt->live_merge_tag_regex, $child_product['name'], $matches );
+
+						$merge_tag_match_value = GFCommon::replace_variables( $matches[1], $nested_form, $child_entry, false, false, false );
+						$child_product['name'] = str_replace( $matches[0], $merge_tag_match_value, $child_product['name'] );
+					}
+
```
claygriffiths, line 55:
```suggestion
					// If Nested Form fields have Live Merge Tags, process those.
					if ( method_exists( 'GP_Populate_Anything_Live_Merge_Tags', 'has_live_merge_tag' ) ) {
						$gppa_lmt         = GP_Populate_Anything_Live_Merge_Tags::get_instance();
				
						if ( $gppa_lmt->has_live_merge_tag( $child_product['name'] ) ) {
							$child_product['name'] = $gppa_lmt->replace_live_merge_tags_static( $child_product['name'], GFAPI::get_form( $field->formId ), $child_entry );
						}
					}
```

claygriffiths: This is untested, but we have some nice methods built in that might be able to simplify this.

claygriffiths (commented):
Nice work! I left a suggested change to simplify things and rely more on some of the methods we already have.

I didn't actually test the change, but it should point you in the right direction.

gp-nested-forms/gpnf-child-products-in-parent-order-summary.php
saifsultanc: The simplified logic does make sense, but ain't work. I am confused about 1 particular method: 

`public function get_live_merge_tag_value( $merge_tag, $form, $entry_values = null ) {` ([L831](https://github.com/gravitywiz/gp-populate-anything/blob/master/includes/class-live-merge-tags.php#L831))

what does the 3rd parameter `$entry_values` imply? Is it the `$entry` object or something else? This method gets called in the call stack via `replace_live_merge_tags_static` or `replace_live_merge_tags`

claygriffiths: Ah, my bad. I forgot to populate the LMT whitelist (it's for security, so we don't just parse arbitrary Live Merge Tags entered by users) and also I provided the wrong form param.

`$entry_values` is _usually_ interchangeable with `$field_values` or `$entry` in the scope of GPPA unless something is explicitly being done to/with the entry.

The following should work:

```php
					// If Nested Form fields have Live Merge Tags, process those.
					if ( method_exists( 'GP_Populate_Anything_Live_Merge_Tags', 'has_live_merge_tag' ) ) {
						$gppa_lmt = GP_Populate_Anything_Live_Merge_Tags::get_instance();

						if ( $gppa_lmt->has_live_merge_tag( $child_product['name'] ) ) {
							$gppa_lmt->populate_lmt_whitelist( $child_form );
							$child_product['name'] = $gppa_lmt->replace_live_merge_tags_static( $child_product['name'], $child_form, $child_entry );
						}
					}
```

saifsultanc: @claygriffiths Form param was fine, `$field->formId` points to the correct nested form. I had it on my snippet as well. Infact `$child_form` is not pointing correctly.

Adding the LMT whitelist to the current nested form, but that still doesn't resolve the issue. The manual approach of calling GFCommon::replace_variables is the only one that I've working thus far.

Out of curiosity, we do have a filter added  (L81):
`add_filter( 'gpnf_all_entries_nested_entry_markup', array( $this, 'replace_live_merge_tags_gpnf_all_entries' ), 10, 5 );`

The implementation is 

```
	function replace_live_merge_tags_gpnf_all_entries( $markup, $nested_form_field, $nested_form, $entry, $args ) {
		$this->populate_lmt_whitelist( $nested_form );
		return $this->replace_live_merge_tags_static( $markup, $nested_form, $entry );
	}
```

On debugging, I can confirm that the filter does run, and nested form and entry values are correctly populated. It's again the call to `replace_live_merge_tags_static` that does not fetch the results. I think if we fix the issue here, we will not need to make updates to the snippet?

claygriffiths: @saifsultanc I sent a Loom to you regarding this. Thanks for continuing to dig in!

gp-nested-forms/gpnf-child-products-in-parent-order-summary.php
```diff
@@ -40,6 +40,16 @@
 				$_child_products    = array();
 				foreach ( $child_product_info['products'] as $child_field_id => $child_product ) {
 					$child_product['name'] = "{$product_info['products'][ $field->id ]['name']} — {$child_product['name']}";
+
+					// If Nested Form fields have Live Merge Tags, process those.
+					if ( method_exists( 'GP_Populate_Anything_Live_Merge_Tags', 'has_live_merge_tag' ) ) {
+						$gppa_lmt = GP_Populate_Anything_Live_Merge_Tags::get_instance();
+						$nested_form =  GFAPI::get_form( $field->formId );
```
claygriffiths, line 47: Is this line needed?

claygriffiths (approved):
LGTM! Just one potential line deletion that can be made.

claygriffiths (approved):
LGTM! I think you can squash this into one commit with Interactive Rebase to make Danger happy.

Let me know if you want to work through it together.

PR #554, opened 2022-10-26
`gppa-conditional-logic-for-empty-dynamic-list.js`: Added snippet for empty dynamic list being used for conditional logic.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2047842152/40019?folderId=3808239

## Summary

When the value of a Populate field is set using the filter [gppa_no_choices_value](https://gravitywiz.com/documentation/gppa_no_choices_value/) if there aren't any results, the value isn't detected in Conditional logic. This snippet detects the selected value after the empty load / set via snippet and processes the conditional logic for the field targeting that list.

### DIFF
--- a/gp-populate-anything/gppa-conditional-logic-for-empty-dynamic-list.js
+++ b/gp-populate-anything/gppa-conditional-logic-for-empty-dynamic-list.js
@@ -1,0 +1,17 @@
+/**
+ * Gravity Perks // Populate Anything // Conditional Logic using a field that populates with no result
+ * https://gravitywiz.com/documentation/gravity-forms-populate-anything/
+ *
+ * 1. Install this snippet with our free Custom JavaScript plugin.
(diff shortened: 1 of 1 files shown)


spivurno (approved):
Approved with two tweaks. 

1. Let's rename the snippet to: `gppa-conditioanl-logic-when-no-results.js`
2. Let's move this to the experimental folder.

Thoughts for the future:

- The issue is that the Drop Down is disabled so when Gravity Forms attempts to fetch the value, it returns null even though the No Results option is selected by default.
- If this comes up again, we could explore using the `gform_is_value_match` filter to modify the rule value and return the desired value ourselves.
- This should likely only be used for display only fields (i.e. HTML fields) as determining if the field has no results on submission will be tricky since disabled fields do not submit their values.

PR #561, opened 2022-11-02
`gp-apf-disable-non-numeric-input.js`: Added snippet to disable non-numeric input.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2055597976/40253?folderId=3808239

## Summary

Disables non-numeric input for the Advanced Phone Field.


claygriffiths:
I think this feels more like something that needs to be fixed in the core of GPAPF. Specifically, some way of clearing out the value if validation fails.

This would also prevent characters such as `-` and `+` from being used.

PR #564, opened 2022-11-04
`gw-day-count.php`: Migrated from GitHub Gists.

scyt:
## Context

📓 Notion: https://www.notion.so/gravitywiz/Calculate-Number-of-Days-Between-Two-Dates-7e814573b8254662a70ec790766077a8

## Summary

Migrating from [this Gist](https://gist.github.com/spivurno/4996579#file-gw-gravity-forms-day-count-php).

### DIFF
--- a/gravity-forms/gw-day-count.php
+++ b/gravity-forms/gw-day-count.php
@@ -1,5 +1,10 @@
 <?php
 /**
+ * --- STOP! ---
+ * Get the latest version:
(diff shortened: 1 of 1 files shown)


scyt:
This is failing PHPCS due to using `extract()`. This will require rewrites to pass.

Also, thoughts on adding a deprecated header, [this one](https://github.com/gravitywiz/snippet-library/blob/ab9d4a25fca98cfa88f59c51a389163246797a94/gravity-forms/gw-calc-subtotal.php#L3-L6)?

PR #566, opened 2022-11-05
Create gw-set-product-category.php

zippa71:
There are several drop down fields. Each subsequent one is filled from the parent by Term_ID using populate anything. In the feed settings, when using this code, you can not specify in which categories the product will be recorded.

## Context

### DIFF
--- a/experimental/gw-gfapc-map-multiple-fields-to-taxonomy.php
+++ b/experimental/gw-gfapc-map-multiple-fields-to-taxonomy.php
@@ -1,0 +1,28 @@
+<?php
+/**
+ * Gravity Wiz // Gravity Forms // Map Multiple Fields w/ Advanced Post Creation
+ * https://gravitywiz.com/
+ *
+ * By default, the Advanced Post Creation add-on does not allow you to map multiple fields to a taxonomy nor does it 
+ * allow you to set terms by ID.
+ * 
+ * This snippet allows to specify multiple fields on a form that have been populated with term IDs (we recommend 
+ * [Populate Anything][1] for this) and to set the taxonomy/terms based on those IDs for the generated post.
+ *
+ * [1]: https://gravitywiz.com/documentation/gravity-forms-populate-anything/
+ */
+add_action( 'gform_advancedpostcreation_post_after_creation', function ( $post_id, $feed, $entry, $form ) {
+	
+	// Update "1", "2", "3" to the field IDs that have been populated with terms. Add additional IDs as needed.
+	$term_field_ids = array( 1, 2, 3 );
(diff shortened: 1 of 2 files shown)


spivurno:
Awesome, @zippa71! 

Just want to make sure I understand...

1. You have a handful of Drop Down fields.
2. Each dropdown is populated (via Populate Anything) based on the term ID selected in the previous dropdown.
3. This snippet will ensure that the post is added to all selected categories and not just the subcategory in the last field.
4. This is required because the Advanced Post Creation add-on won't allow you to select more than one field to set terms from for a given taxonomy.

Is that right?

spivurno:
@zippa71 Thanks for the extra information. I've made some changes to the snippet to bring it inline with our Snippet Library standards (which I'm realizing aren't public 😅). Could you let me know if this is still working for you?

zippa71:
> 
`add_action( 'gform_advancedpostcreation_post_after_creation', function ( $post_id, $feed, $entry, $form ) {

	// Update "1", "2", "3" to the field IDs that have been populated with terms. Add additional IDs as needed.
	$term_field_ids = array( 1, 3 );
	// Update "categories" to the name of the taxonomy to which the populated terms belong.
	$taxonomy       = 'product_cat';
	$term_ids = array();
	foreach ( $term_field_ids as $term_field_id ) {
		$term_ids[] = (int)$entry[ $term_field_id ];
	}

	wp_set_object_terms( $post_id, $term_ids, $taxonomy );

}, 10, 4 `);``

zippa71:
> @zippa71 Thanks for the extra information. I've made some changes to the snippet to bring it inline with our Snippet Library standards (

> which I'm realizing aren't public 😅

). Could you let me know if this is still working for you?

I'm not very good at github, but I really want to help those who come across this problem. If you add this code to your collection, I will be very happy

spivurno:
@zippa71 Looks like you may have closed the PR by accident. I've re-opened it and updated the snippet so that the term IDs are passed as integers. Good catch!

With that change, is the new version of this snippet working for you? Planning to merge this into the library as soon as I have your confirmation. 🙂

zippa71:
> @zippa71 Looks like you may have closed the PR by accident. I've re-opened it and updated the snippet so that the term IDs are passed as integers. Good catch!
> 
> With that change, is the new version of this snippet working for you?

>  Planning to merge this into the library as soon as I have your confirmation. 🙂
I confirm!

Good afternoon.
To summarize, this setting is only useful if you are passing field values as term _ID when selecting categories in GP Populate Anything. Then it allows you to find a category by term _ID and write a new product to the right place.

spivurno:
Thanks for your contribution, @zippa71!

PR #572, opened 2022-11-08
`gppa-replace-field-with-spinner-instead-of-pulsing.js`: Added Instruction Video.

levisimon:
## Context
`gppa-replace-field-with-spinner-instead-of-pulsing.js`: Added Instruction Video.


spivurno:
This Loom no longer exists. Closing.

PR #576, opened 2022-11-16
`gw-zip-files.php`: Updated snippet to ensure all nested form entry uploads are added to zip.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2068662957/40696?folderId=3808239

## Summary

Nested form entries have their uploads overridden on `$nested_archive_files` (associative array in play) which causes only the last nested form entry to be sent via zip. This update first adds all entry files at separate array indexes, then when processing archive files, we pluck and only copy the values to avoid `[path]` overrides when prep-ing the archive files array. This ensure all files from each nested form entry are successfully added to the zip

### DIFF
--- a/gravity-forms/gw-zip-files.php
+++ b/gravity-forms/gw-zip-files.php
@@ -79,7 +79,7 @@ class GW_Zip_Files {
 				if ( is_wp_error( $nested_entry ) ) {
 					continue;
 				}
-				$nested_form          = GFAPI::get_form( $nested_entry['form_id'] );
+				$nested_form = GFAPI::get_form( $nested_entry['form_id'] );
 				// Add each nested entry files list as a separate array to avoid overriding previously saved files with the same associative array key (the field id).
 				$nested_archive_files = array_merge( $nested_archive_files, array( $this->get_entry_files( $nested_entry, $nested_form ) ) );
 			}
--- a/gravity-forms/gw-zip-files.php
+++ b/gravity-forms/gw-zip-files.php
@@ -80,17 +80,23 @@ class GW_Zip_Files {
 					continue;
 				}
 				$nested_form          = GFAPI::get_form( $nested_entry['form_id'] );
-				$nested_archive_files = array_merge( $nested_archive_files, $this->get_entry_files( $nested_entry, $nested_form ) );
+				// Add each nested entry files list as a separate array to avoid overriding previously saved files with the same associative array key (the field id).
+				$nested_archive_files = array_merge( $nested_archive_files, array( $this->get_entry_files( $nested_entry, $nested_form ) ) );
 			}
 		}
 
-		$archive_files = $this->get_entry_files( $entry, $form );
+		// For conformity with the $nested_archive_files, the current (parent) entry files are also loaded onto an array.
+		$archive_files = array( $this->get_entry_files( $entry, $form ) );
 		$archive_files = array_merge( $archive_files, $nested_archive_files );
 		if ( empty( $archive_files ) ) {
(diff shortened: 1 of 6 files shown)


gravity-forms/gw-zip-files.php
```diff
@@ -80,7 +80,7 @@ public function archive_files( $entry, $form ) {
 					continue;
 				}
 				$nested_form          = GFAPI::get_form( $nested_entry['form_id'] );
-				$nested_archive_files = array_merge( $nested_archive_files, $this->get_entry_files( $nested_entry, $nested_form ) );
+				$nested_archive_files = array_merge( $nested_archive_files, array( $this->get_entry_files( $nested_entry, $nested_form ) ) );
```
claygriffiths, line 83: Curious, why did this need to be wrapped in an array? What is the resulting structure of `$nested_archive_files`?

gravity-forms/gw-zip-files.php
```diff
@@ -90,7 +90,11 @@ public function archive_files( $entry, $form ) {
 			return;
 		}
 
-		$archive_file_paths = wp_list_pluck( $archive_files, 'path' );
+		$archive_file_paths = array();
+		foreach ( $archive_files as $archive_file ) {
+			$archive_file_path  = array_values( wp_list_pluck( $archive_file, 'path' ) );
+			$archive_file_paths = array_merge( $archive_file_paths, $archive_file_path );
+		}
```
claygriffiths, line 97: Does this still work with non-GPNF fields?

claygriffiths (commented):
Good stuff! Just a couple of questions from me.

gravity-forms/gw-zip-files.php
saifsultanc: @claygriffiths  I wrapped in an array, so previous nested form entry files are not overridden. This loads in associative array form, so if the file upload field on the nested form is field ID 3. Let's assume we have uploaded 2 files, so we have a 3.1 and a 3.2 on the associative array. When we process the next nested form entry, the 3.1, 3.2 (or whatever the number of files uploaded for the second nested entry) will be updated to the current array thus losing the previous entry file links. I just felt it was more of a logical update, then we can run through each array index and extract the files of each entry (as many as there are). Let me know if you want to see a screencast with a debugger

gravity-forms/gw-zip-files.php
saifsultanc: @claygriffiths Thanks for that. I updated L87 to load `$archive_files = array( $this->get_entry_files( $entry, $form ) );` in the same "array way". Then the merge goes smoothly. So basically I can confirm I have tested:

1. All nested form entries combined zip.
2. Files from nested form entries and parent entry in a combined zip.
3. No nested form entry, only parent form entry zip.

Anything I am missing still?

gravity-forms/gw-zip-files.php
claygriffiths: Got it! Good idea.

gravity-forms/gw-zip-files.php
claygriffiths: Nice work! This makes sense.

It does feel a bit weird that any calls to `get_entry_files` get wrapped in another array so I'd definitely recommend adding some inline comments to this. Pretty much everything you mentioned in these comments was super helpful so if you could distill it into inline/block comments that'd be perfect.

saifsultanc:
@claygriffiths Added inline comments. Please let me know if it looks good to merge?

PR #577, opened 2022-11-17
`gw-multi-file-merge-tag.php`: Added new "markup/container" property for specifying container markup that will contain all file markup.

spivurno:
- Added new "markup/container" property for specifying container markup that will contain all file markup.
- Added `gwmfmt_tags` filter for creating custom tags (like {filesize}). 
- Fixed issue in default markup.

## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2065421606/40617/

## Summary

Customer wanted to more easily define a container as part of her Multi-file Merge Tag template and also to include the file's file size in the generated markup.

This snippet uses the new `gwmfmt_tags` filter to add a custom tag for {filesize} which renders the filesize in kb's.

```php
add_filter( 'gwmfmt_tags_123', function( $tags, $file ) {
	$tags['{filesize}'] = round( filesize( GFFormsModel::get_physical_file_path( $file ) ) / 1000 ) . 'KB';
	return $tags;
}, 10, 2 );
```

Example of the configuration that uses both new options:

```php
gw_multi_file_merge_tag()->register_settings( array(
	'form_id' => 2,
	'field_ids' => array( 4 ),
	'markup' => array(
		'container' => '<ul>%s</ul>',
		array(
			'file_types' => array( 'jpg', 'jpeg', 'png', 'gif' ),
			'markup'     => ' <li class="gw-file gw-image gw-{ext}" style="width: 100px; height: 100px;"> <a href="{url}"><img src="{url}"></a> <a href="{url}">{filename} ({filesize})</a> </li> '
		)
	)
) );
```

### DIFF
--- a/gravity-forms/gw-multi-file-merge-tag.php
+++ b/gravity-forms/gw-multi-file-merge-tag.php
@@ -230,12 +230,12 @@ class GW_Multi_File_Merge_Tag {
 			$markup_found = true;
 			$markup       = $file_type_markup['markup'];
 
-			$tags = array(
+			$tags = gf_apply_filters( array( 'gwmfmt_tags', $form_id ), array(
 				'{url}'      => $file,
 				'{filename}' => $filename,
 				'{basename}' => $basename,
 				'{ext}'      => $extension,
-			);
+			), $file, $file_info, $form_id );
 
 			foreach ( $tags as $tag => $tag_value ) {
 				$markup = str_replace( $tag, $tag_value, $markup );
--- a/gravity-forms/gw-multi-file-merge-tag.php
+++ b/gravity-forms/gw-multi-file-merge-tag.php
@@ -13,7 +13,7 @@
  * Plugin URI:   https://gravitywiz.com/customizing-multi-file-merge-tag/
  * Description:  Enhance the merge tag for multi-file upload fields by adding support for outputting markup that corresponds to the uploaded file.
  * Author:       Gravity Wiz
- * Version:      1.7.5
+ * Version:      1.8
  * Author URI:   https://gravitywiz.com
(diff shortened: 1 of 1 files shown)


gravity-forms/gw-multi-file-merge-tag.php
```diff
@@ -48,9 +48,10 @@ public function get_default_args() {
 			'form_id'        => false,
 			'field_ids'      => array(),
 			'exclude_forms'  => array(),
-			'default_markup' => '<li><a href="{url}">{filename}.{ext}</a></li>',
+			'default_markup' => '<a href="{url}">{filename}.{ext}</a>',
```
claygriffiths, line 51: Do we want to no longer provide this at all as a default?

claygriffiths (approved):
Good stuff!

I have one question and PHPCS is barking about something.

claygriffiths:
@spivurno I updated the commits and PR title to make Danger happy 😃

gravity-forms/gw-multi-file-merge-tag.php
spivurno: I removed the <li> wrapper from the default markup as a I realized we do not output the files in a <ul> by default and none of the file-type-specific default markup used <li>'s either.

There's an argument that we should default to <li>'s for everything since GF's default output for Multi File merge tags is wrapped in <ul> and <li>'s but given the lack of customer feedback about this during the lifetime of the snippet, I went with the least intrusive adjustment.

Thoughts?

claygriffiths: That works for me! We can always adjust.

PR #578, opened 2022-11-18
`gpep-edit-entry.php`: Fixed snippet to update entry on submit.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2057441367/40322?folderId=7098280#thread-6158158999

## Summary

The conditional field if it got hidden/removed due to an update in the entry on Edit Entry had it's value still retained. The entry object was getting updated to remove the value. This update fires a `update_entry` on submit, to ensure the entry is completely updated.

### DIFF
--- a/gp-easy-passthrough/gpep-edit-entry.php
+++ b/gp-easy-passthrough/gpep-edit-entry.php
@@ -35,6 +35,8 @@ class GPEP_Edit_Entry {
 		// Bypass limit submissions on validation
 		add_filter( 'gform_validation', array( $this, 'bypass_limit_submission_validation' ) );
 
+		// Update entry values on submission
+		add_action( 'gform_after_submission', array( $this, 'gpep_update_entry' ), 10, 2 );
 	}
 
 	public function capture_passed_through_entry_ids( $form, $values, $passed_through_entries ) {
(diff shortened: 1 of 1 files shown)


gp-easy-passthrough/gpep-edit-entry.php
```diff
@@ -35,6 +35,8 @@ public function __construct( $options ) {
 		// Bypass limit submissions on validation
 		add_filter( 'gform_validation', array( $this, 'bypass_limit_submission_validation' ) );
 
+		// Update entry values on submission
+		add_action( 'gform_after_submission', array( $this, 'gpep_update_entry' ), 10, 2 );
```
claygriffiths, line 39: Do you know the reason technically why this is needed? It should already be updating the entry thanks to https://github.com/gravitywiz/snippet-library/pull/578/files#diff-e8542d5676691b24606b326cc6bc02c9ac4c7990632c02bb5133f709918fd0f7R30

If this is needed, we'll want to make sure it's scoped to this particular form.

saifsultanc: The link just points me back to the git compare of this branch. From what I recall, the entry data wasn't matching post edit, which is where the idea of "edit entry" firing GFAPI update came.

saifsultanc:
Fixed with another PR

PR #581, opened 2022-11-23
`gw-zip-files.php`: Updated snippet to ensure zip has renamed file names done with GP File Renamer.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2068662957/40696?folderId=3808239

## Summary

The `archive_files` method for the zip files snippet was hooked on `gform_entry_created`. The file renamer does the file renaming hooked at `gform_entry_post_save`. While the file entry still got the nested form renamed file pattern (entry save processed), the parent form did not because the zip was getting created before the file renamer operation was executed. The update here just makes the `archive_files` method execute on the same hook as the file rename operation but slightly later. The File Renamer has its filter hook call at priority 9, the snippet here does it at priority 10.

### DIFF
--- a/gravity-forms/gw-zip-files.php
+++ b/gravity-forms/gw-zip-files.php
@@ -43,7 +43,7 @@ class GW_Zip_Files {
 		add_filter( 'gform_entry_meta', array( $this, 'register_entry_meta' ), 10, 2 );
 		add_filter( 'gform_entries_field_value', array( $this, 'modify_zip_display_value' ), 10, 3 );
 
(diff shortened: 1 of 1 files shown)


claygriffiths (approved):
LGTM! I assume this works without File Renamer enabled, too?

saifsultanc:
> 



> LGTM! I assume this works without File Renamer enabled, too?

Yes! Working fine with (and without) File Renamer.

PR #584, opened 2022-11-29
`gw-populate-date.php`: Updated snippet to fix the hour 12 not updating to PM with arithmetic.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2076433976/41028?folderId=3808239

## Summary

When the desired populate date arithmetic was computing Hours 1 and onwards, the AM to PM conversion happened fine. The only issue was with Hour 12. This PR adds a condition to the logic to check the hour "12" when AM to PM swap should occur.


claygriffiths (approved):
LGTM!

PR #586, opened 2022-12-02
`gw-choice-counter.php`: Updated snippet with a values parameter to count checkboxes that match the values given.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2083255684/41352?folderId=3808239

## Summary

The 'values' parameter added. This can be passed as an array of values, and the checkboxes will only be counted if values match that array.

Here is screencast: https://www.loom.com/share/7b8e9acd886247b8887ec56022ce363a

### DIFF
--- a/gravity-forms/gw-choice-counter.php
+++ b/gravity-forms/gw-choice-counter.php
@@ -116,7 +116,7 @@ class GW_Choice_Count {
 						for ( var i = 0; i < choiceFieldIds.length; i++ ) {
 
 							var $choiceField = $( '#input_' + formId + '_' + choiceFieldIds[ i ] );
-							if ( values.length == 0) {
+							if ( ! values ) {
 								// If no values provided in the config, just get the number of checkboxes checked.
 								if ( self.isCheckboxField( $choiceField ) ) {
 									count += $choiceField.find( 'input[type="checkbox"]:checked' ).not(' #choice_' + choiceFieldIds[ i ] + '_select_all').length;


spivurno (changes requested):
Looking good! 

One functional change. Any matched value should count. So if you have checkboxes for: 

- [x] First Choice
- [ ] Second Choice
- [ ] Third Choice
- [x] Fourth Choice

And the "values" parameter is set to `array( 'First Choice', 'Second Choice' )`, the count would be `1` because "First Choice" is checked.

- [x] First Choice
- [x] Second Choice
- [ ] Third Choice
- [x] Fourth Choice

This configuration would give a count of `2` because "First Choice" and "Second Choice" are checked.

saifsultanc:
> Looking good!
> 
> One functional change. Any matched value should count. So if you have checkboxes for:
> 
> * [x]  First Choice
> * [ ]  Second Choice
> * [ ]  Third Choice
> * [x]  Fourth Choice
> 
> And the "values" parameter is set to `array( 'First Choice', 'Second Choice' )`, the count would be `1` because "First Choice" is checked.
> 
> * [x]  First Choice
> * [x]  Second Choice
> * [ ]  Third Choice
> * [x]  Fourth Choice
> 
> This configuration would give a count of `2` because "First Choice" and "Second Choice" are checked.

As mentioned, updated to 'Any' instead of 'All'. Thanks @spivurno !

gravity-forms/gw-choice-counter.php
```diff
@@ -107,18 +108,28 @@ function output_script() {
...
 							} else {
-								count += $choiceField.find( 'option:selected' ).length;
+								var selectedValues = [];
+								$choiceField.find( 'input[type="checkbox"]:checked' ).each( function( k, $selectedChoice ) {
+									selectedValues.push( $selectedChoice.value );
+								});
+								values.forEach( function( val ) {
+									count += selectedValues.indexOf( val ) >= 0;
+								});
 							}
```
claygriffiths, line 133: Can you add some inline comments here of what's going on?

claygriffiths (commented):
Deferring to @spivurno for final review on this one. Otherwise, LGTM on first glance!

gravity-forms/gw-choice-counter.php
saifsultanc: Added

spivurno (changes requested):
This doesn't appear to work if no "values" parameter is passed. I would expect all checkboxes to count in that case.

```
new GW_Choice_Count( array(
	'form_id'          => 595,
	'count_field_id'   => 4,
	'choice_field_ids' => array( 1, 3 ),
) );
```

spivurno (approved):
Working well! ✅

PR #587, opened 2022-12-02
`gw-require-unique-values.php`: Updated snippet to support checking over unique values on columns for List field type.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2084259592/41399?folderId=3808239

## Summary

By default, the snippet is designed to compare the values in two different fields and confirm that they are different.

This alters the behavior if a single field ID is passed and that field is a List field, we would process it for unique values over columns. For example,

A | B | C
A | D | E
Fails. A is present multiple times in the first column.

A | B | C
D | A | E
Passes. A is present multiple times but in different columns.

### DIFF
--- a/gravity-forms/gw-require-unique-values.php
+++ b/gravity-forms/gw-require-unique-values.php
@@ -29,7 +29,7 @@ class GW_Require_Unique_Values {
 		) );
 
 		// No validation when the field id list is empty.
-		if ( ! $this->_args['field_ids'] ) {
+		if ( empty( $this->_args['field_ids'] ) ) {
 			return;
 		}
 
@@ -99,9 +99,9 @@ class GW_Require_Unique_Values {
 		$is_unique = true;
 
 		// Multi-Column List
-		if ( is_array( $value[0] ) ) {
+		if ( is_array( rgar( $value, 0 ) ) ) {
 			foreach ( array_keys( $value[0] ) as $column_name ) {
-				$column_values = array_column( $value, $column_name );
+				$column_values = wp_list_pluck( $value, $column_name );
 				// Count unique values in each column.
 				if ( $rows != count( array_unique( $column_values ) ) ) {
 					$is_unique = false;
--- a/gravity-forms/gw-require-unique-values.php
+++ b/gravity-forms/gw-require-unique-values.php
@@ -28,6 +28,11 @@ class GW_Require_Unique_Values {
 			'case_sensitive'      => false,
 		) );
 
+		// No validation when the field id list is empty.
+		if ( ! $this->_args['field_ids'] ) {
+			return;
+		}
+
 		$this->_args['master_field_id'] = array_shift( $this->_args['field_ids'] );
 
 		// do version check in the init to make sure if GF is going to be loaded, it is already loaded
@@ -111,7 +116,7 @@ class GW_Require_Unique_Values {
 
 		if ( $result['is_valid'] && ! $is_unique ) {
 			$result['is_valid'] = false;
-			$result['message']  = $this->_args['validation_message'] . ' All columns should have unique values.';
+			$result['message']  = $this->_args['validation_message'];
 		}
 
 		return $result;
@@ -186,7 +191,7 @@ class GW_Require_Unique_Values {
 	public function is_applicable_field( $field ) {
 
 		// A single list field can be used for validation (validation logic to use unique values check over columns).
-		if ( ! $this->_args['field_ids'] && $field->get_input_type() == 'list' ) {
+		if ( $field->id == $this->_args['master_field_id'] && ! $this->_args['field_ids'] && $field->get_input_type() == 'list' ) {
 			return true;
 		} elseif ( ! $this->_args['field_ids'] ) {
 			return false;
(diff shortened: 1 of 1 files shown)


spivurno (changes requested):
Multi-column List fields are working well!

Single-column List fields are generating some errors and not validating as expected.

CleanShot 2022-12-02 at 11 56 02

Form export: https://gwiz.io/3unrJiR

Configuration:

```php
new GW_Require_Unique_Values( array(
	'form_id'   => 592,
	'field_ids' => array( 1 ),
) );

new GW_Require_Unique_Values( array(
	'form_id'   => 592,
	'field_ids' => array( 3 ),
) );
```

saifsultanc:
> Multi-column List fields are working well!
> 
> Single-column List fields are generating some errors and not validating as expected.
> 
> CleanShot 2022-12-02 at 11 56 02
> 
> Form export: https://gwiz.io/3unrJiR
> 
> Configuration:
> 
> ```
> new GW_Require_Unique_Values( array(
> 	'form_id'   => 592,
> 	'field_ids' => array( 1 ),
> ) );
> 
> new GW_Require_Unique_Values( array(
> 	'form_id'   => 592,
> 	'field_ids' => array( 3 ),
> ) );
> ```

@spivurno Thanks, David. Snippet fixed!

gravity-forms/gw-require-unique-values.php
```diff
@@ -85,6 +89,34 @@ public function validate( $result, $value, $form, $field ) {
...
+		} else {
+			// Count unique values in the Single Column List.
+			if ( $rows != count( array_unique( $value ) ) ) {
+				$is_unique = false;
+			}
+		}
+
+		if ( $result['is_valid'] && ! $is_unique ) {
+			$result['is_valid'] = false;
+			$result['message']  = $this->_args['validation_message'] . ' All columns should have unique values.';
```
spivurno, line 114: Like the idea of providing additional clarification about the validation error; however, the user already has the ability to define their own validation message and this would prevent them from customizing that in its entirety. For simplicity, I would simply use the default validation message that the user can modify.

gravity-forms/gw-require-unique-values.php
```diff
@@ -153,7 +185,10 @@ public function get_value_hash( $value ) {
 
 	public function is_applicable_field( $field ) {
 
-		if ( ! $this->_args['field_ids'] ) {
+		// A single list field can be used for validation (validation logic to use unique values check over columns).
+		if ( ! $this->_args['field_ids'] && $field->get_input_type() == 'list' ) {
```
spivurno, line 189: The user should still be required to pass the List field ID as the `field_ids` parameter. The way this is written will make this apply to all List fields if no field IDs are passed. For example, this configuration shouldn't validate anything:

```
new GW_Require_Unique_Values( array(
	'form_id'   => 592,
	'field_ids' => array(),
) );
```

...but it will currently validate all List fields.

spivurno (changes requested):
A few more tweaks and I think we're there!

gravity-forms/gw-require-unique-values.php
saifsultanc: Got it. Updated

gravity-forms/gw-require-unique-values.php
saifsultanc: Got it. Updated the condition here. Also, add a check on the constructor side to completely skip the validation if the field id (coming via the args) is empty. Thanks, @spivurno!

gravity-forms/gw-require-unique-values.php
```diff
@@ -28,6 +28,11 @@ public function __construct( $args = array() ) {
 			'case_sensitive'      => false,
 		) );
 
+		// No validation when the field id list is empty.
+		if ( ! $this->_args['field_ids'] ) {
```
claygriffiths, line 32: I'd probably go with `empty()` here

gravity-forms/gw-require-unique-values.php
```diff
@@ -85,6 +94,34 @@ public function validate( $result, $value, $form, $field ) {
 		return $result;
 	}
 
+	public function validate_list( $result, $value ) {
+		$rows      = count( $value );
+		$is_unique = true;
+
+		// Multi-Column List
+		if ( is_array( $value[0] ) ) {
```
claygriffiths, line 102: Will this throw errors if `$value` is empty? I think you could use `rgar( $value, 0 )` if it is.

gravity-forms/gw-require-unique-values.php
```diff
@@ -85,6 +94,34 @@ public function validate( $result, $value, $form, $field ) {
...
+			foreach ( array_keys( $value[0] ) as $column_name ) {
+				$column_values = array_column( $value, $column_name );
```
claygriffiths, line 104: TIL about `array_column`! We've always used `wp_list_pluck()`.

claygriffiths (commented):
Deferring to @spivurno for final review.

spivurno (approved):
Perfect! ✅

PR #588, opened 2022-12-06
`gpcc-selectively-copy-based-on-radio-choice.js`: Added a snippet toselectively copy cat values.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2085288395/41436?folderId=3808239

https://www.loom.com/share/93e760f1004d402289151c57af780889

## Summary

The source field for the copy cat operation is static by default, this snippet gives the capability of dynamically selecting the source field over a radio button choice list. Ensure the field IDs are stored on the radio button values, the ones which need to be targetted for copying.


spivurno (approved):
I'm not sure that we can rely on the value of the choice to be the indicator for which value can be copied but feel free to share this with the customer and get their feedback. 

Other than that concern, this looks great!

PR #589, opened 2022-12-06
`gpns-gravityview-notifications-on-approve.php`: Added a snippet for GravityView Notification Trigger.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2083497762/41364?folderId=3808239

## Summary

Using GPNS with GravityView's Approve/Disapprove is causing notifications to not send if the notification delay passes before the entry is approved. This snippet approach aims at overriding the notification delay timestamp comparison.


claygriffiths (approved):
LGTM pending PHPCS fixes

PR #591, opened 2022-12-08
`gpld-last-day-of-the-month-only.php`: Added snippet to display last day of the month only for GPLD.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2090349115/41601?folderId=3808239

## Summary

This snippet only displays the last day of each month for the GPLD.


claygriffiths (changes requested):
I think this could be simplified to use `strtotime` like so:

```php
date('m/d/Y', strtotime('last day of this month'));
```

Also, there are a few PHPCS warnings/errors that can be cleaned up.

PR #596, opened 2022-12-21
`gpfr-allow-commas.php`: Added initial support.

veryspry:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2103452211/42095/ 

## Summary

* Relies on a new, unreleased filter in GPFR called `gpfr_sanitize_file_name_chars`.
* Removes commas from the list of dissallowed characters when renaming files with GPFR.

### DIFF
--- a/gp-file-renamer/gpfr-allow-commas.php
+++ b/gp-file-renamer/gpfr-allow-commas.php
@@ -1,0 +1,20 @@
+<?php
+
+/**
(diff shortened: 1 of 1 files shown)


claygriffiths (commented):
LGTM! 

Formatting is wonky (line length limit being hit, perhaps?), but if PHPCS is happy, I am too 😂

PR #601, opened 2023-01-09
`gw-submit-to-access.php`: Added `cookie_expiration` parameter to allow setting a custom expiration for the cookie that tracks which forms have been submitted.

spivurno:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2114674920/42374

## Summary

Customer required a way to expire the cookie at midnight every night. The new `cookie_expiration` parameter can be configured to do so with the following:

```php
if ( is_callable( 'gw_submit_to_access' ) ) {
	gw_submit_to_access( array(
		'is_persistent' => false,
		'cookie_expiration' => strtotime( 'midnight tomorrow' ),
	) );
}
```

### DIFF
--- a/gravity-forms/gw-submit-to-access.php
+++ b/gravity-forms/gw-submit-to-access.php
@@ -8,7 +8,7 @@
  * Plugin URI:  https://gravitywiz.com/submit-gravity-form-access-content/
  * Description: Require that a form be submitted before a post or page can be accessed.
  * Author:      Gravity Wiz
- * Version:     1.11
+ * Version:     1.12
  * Author URI:  https://gravitywiz.com
  */
 class GW_Submit_Access {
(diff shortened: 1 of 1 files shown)


gravity-forms/gw-submit-to-access.php
```diff
@@ -355,7 +356,7 @@ public function add_submitted_form( $form ) {
 			if ( $this->_args['enable_user_meta'] && is_user_logged_in() ) {
 				update_user_meta( get_current_user_id(), 'gwsa_submitted_forms', $submitted_forms );
 			} else {
-				$expiration = $this->_args['is_persistent'] ? strtotime( '+1 year' ) : null;
+				$expiration = $this->_args['is_persistent'] ? strtotime( '+1 year' ) : $this->_args['cookie_expiration'];
```
claygriffiths, line 359: Should this be something like the following?

```suggestion
				$expiration = $this->_args['is_persistent'] ? rgar( $this->_args, 'cookie_expiration', strtotime( '+1 year' ) ) : null;
```

If it's _not_ persistent, it should always be null, right?

spivurno: That wasn't exactly the original intent of the "is_persistent" parameter but I like making it so moving forward.

claygriffiths: @spivurno Can you elaborate? Isn't the idea that if it's not persistent it's only for the session?

claygriffiths:
LGTM! I recommend Squash & Merge since it added a commit from the suggestion.

PR #603, opened 2023-01-10
`gw-choice-counter.php`: Updated snippet to tackle count field issue when hidden via conditional logic.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2120099163/42554/

## Summary

If hidden via conditional logic, the count field when unhidden would display empty instead of 0. This update overrides that. The rest of the functionality works normally.

### DIFF
--- a/gravity-forms/gw-choice-counter.php
+++ b/gravity-forms/gw-choice-counter.php
@@ -113,31 +113,34 @@ class GW_Choice_Count {
 						var countField = $( '#input_' + formId + '_' + countFieldId ),
 							count      = 0;
 
-						for ( var i = 0; i < choiceFieldIds.length; i++ ) {
+						// Prevent count field from being recalculated if it's hidden.
+						if ( ! gformIsHidden( countField ) ) {
+							for ( var i = 0; i < choiceFieldIds.length; i++ ) {
 
-							var $choiceField = $( '#input_' + formId + '_' + choiceFieldIds[ i ] );
-							if ( ! values ) {
-								// If no values provided in the config, just get the number of checkboxes checked.
-								if ( self.isCheckableField( $choiceField ) ) {
-									count += $choiceField.find( ':checked' ).not(' #choice_' + choiceFieldIds[ i ] + '_select_all').length;
+								var $choiceField = $( '#input_' + formId + '_' + choiceFieldIds[ i ] );
+								if ( ! values ) {
+									// If no values provided in the config, just get the number of checkboxes checked.
+									if ( self.isCheckableField( $choiceField ) ) {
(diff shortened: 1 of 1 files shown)


veryspry (approved):
This makes sense to me!

spivurno (changes requested):
@saifsultanc Great idea! But it does have a fatal flaw. Recorded [a video](https://www.loom.com/share/d98e9e4359d34784ba7d28eca9f26470) and [a form export](https://gwiz.io/3W6mO0V).

As you'll see in the video we need to account for two things:

1. Ensure that the count field is not updated if the count field is hidden via conditional logic. 
2. Ensure that the count field is recalculated when revealed by conditional logic.

For the latter, I would recommend checking out the `gform_post_conditional_logic_field_action` JS action.

saifsultanc:
@spivurno Thanks David! Have pushed the updates, and here is my walkthrough of it : https://www.loom.com/share/3b4db5aeb9ea49989943c3265ba4dc41

Lmk what you think.

gravity-forms/gw-choice-counter.php
```diff
@@ -113,34 +113,59 @@ function output_script() {
...
+					gform.addAction('gform_post_conditional_logic_field_action', function (formId, action, targetId, defaultValues, isInit) {
+						var id = targetId.split( '_' ).pop();
+						if ( id == args.countFieldId ) { 
+							if ( action == 'hide' ) {
+								$( targetId ).find( 'input' ).val('0').trigger( 'change' );
+							} else if ( action == 'show' ) {
+								self.updateChoiceEventHandler();
+							}
+						}
+					});
```
spivurno, line 169: My instinct is to simplify this code to this:

```js
// Recalculate Count field when it is revealed by conditional logic.
gform.addAction( 'gform_post_conditional_logic_field_action', function ( formId, action, targetId ) {
	var id = targetId.split( '_' ).pop();
	if ( id == args.countFieldId && action === 'show' ) {
		self.updateChoiceEventHandler();
	}
} );
```

We don't necessarily need the Count field to be `0` when hidden via conditional logic and since we haven't encountered a use-case that requires it, letting GF clear the value when hidden works well.

spivurno (approved):
Approved with one suggested tweak. 🤘

gravity-forms/gw-choice-counter.php
saifsultanc: @spivurno skipping the '0' update on the counter field doesn't completely work. Here is a quick 40s video of how it is after the suggested update: https://www.loom.com/share/fe4ebb34e33f4580b264670899f591ec
​
​Lmk what you think!

saifsultanc: @spivurno Updated!

PR #608, opened 2023-01-20
`gw-multiple-entries-list-field.php`: Updated snippet to send notifications for created entries.

saifsultanc:
# Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2128964925/42994?folderId=7098280

## Summary

The snippet creates entries from the list fields as per the field mapping rules defined. However, any notifications assigned under Gravity Forms Settings are not fired. This snippet adds the call to Gravity Forms notifications to fire them if they are defined under settings.


spivurno:
Good stuff, @saifsultanc!

Rather than sending a notification for each entry by default, we should probably allow the user to enable this via a parameter. Something like:

```
'send_notifications' => false,
```

claygriffiths (commented):
LGTM pending David's request!

PR #609, opened 2023-01-22
Alter validation and code example at bottom to support the 2 state abbr

ksorensen76:
Alter validation and code example at bottom to support the 2 state abbreviation as key such as in example 1 of gform_us_states (https://docs.gravityforms.com/gform_us_states/).


spivurno:
Thanks for your contribution, @ksorensen76!

PR #610, opened 2023-01-27
`gpui-sequential-ids-as-per-date-field.php`: Added snippet to generate Unique IDs as per selected date field.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2134201637/43258?folderId=3808239#thread-6361627895

## Summary

Filter to generate Unique IDs as per the selected date field.


spivurno (changes requested):
This will need to be tweaked based on the feedback in https://github.com/gravitywiz/gp-unique-id/pull/23.

Looking good though!

PR #613, opened 2023-01-30
`gw-populate-date.php`: Improved time calculations on the current date by using the current time (adjusted for server offset) instead of midnight; also addressed issue with time math on 12PM.

claygriffiths:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2085621576/41445?folderId=1530128

## Summary

Instead of using midnight for the current date when performing time calculation on it (example: `-1 hour`), use the current time adjusted for the server timezone.

Now, when doing `-1 hour` when the current server datetime is Jan 30th 2:45 PM will result in Jan 30th 1:45 PM instead of Jan 29th 11:00 PM.

Additionally, there was an issue where 12:00 PM was being converted to 12:00 AM due to the AM/PM logic not being robust enough in a method.


spivurno (approved):
Good stuff, dude!

PR #614, opened 2023-02-01
Update post date with either a date or time or both.

sbassah:
## Context
https://www.notion.so/gravitywiz/Update-How-to-Update-Posts-Update-Posts-Date-and-Time-2c6804a35da642eba85b49ffaec5b2aa

Added support to set the post date with a single value like 11 which can be a Date or Time field ID. If it is a Date field, the time is set to midnight. If it is a Time field, the date is set to the current date. 

new GW_Update_Posts( array(
	'form_id' => 123,
	'post_id' => 1,
	'post_date'  => 11, // This can be either the Date or Time field ID.
) );

It also supports setting the Post date with an array of Date and Time field IDs like so

new GW_Update_Posts( array(  
        'form_id' => 123,  
        'post_id' => 1,  
        'post_date' => array(  
                        'date' => 11,  
                        'time' => 12,  
        ),  
) );

### DIFF
--- a/gravity-forms/gw-update-posts.php
+++ b/gravity-forms/gw-update-posts.php
@@ -97,7 +97,7 @@ class GW_Update_Posts {
 			$time = $this->_args['date_time']['time'];
 
 			$date = rgar( $entry, $date );
-			$time = rgar( $entry, $time );
+			$time = rgar( $entry, $time, '00:00 am' );
 
 			if ( empty( $date ) ) {
 				$date = explode( ' ', $post->post_date )[0];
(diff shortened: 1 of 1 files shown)


spivurno (approved):
@sbassah Great work, man! I moved the logic to its own function since it was getting beefy and tweaked it slightly for clarity. Could you test this one more time on your end and make sure all the scenarios are still working?

Exciting to see you producing code of this quality!

sbassah:
@spivurno  It works when the array parameter is used, but it throws a fatal error with the message below when the integer parameter is used. 

`Fatal error: Uncaught Error: Cannot use a scalar value as an array in...`

PR #620, opened 2023-02-13
`gw-force-greater-end-time.js`: Added new snippet for Gravity Forms that forces users to enter a greater end time.

sbassah:
## Context

When a form has a start time field and end time field, whereby the end time is supposed to be greater than the start time, this 
snippet populates the end time field with a time one hour ahead of the start time value. Also, it prevents the user from entering a time in the past. 

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2152005323/44085?folderId=3775888


spivurno (approved):
✅✅✅

PR #624, opened 2023-02-15
`gpasc-modify-drafts-shortcode-user-id.php`: Added snippet to allow modification of `[gpasc_drafts]` `user_id` arg.

veryspry:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2145965518/43807/

## Summary

This allows modification of the `user_id` arg for the `[gpasc_drafts]` shortcode.

This arg is currently not allowed to be set via standard syntax and must be set via filter as demonstrated here.

### DIFF
--- a/gp-advanced-save-and-continue/gpasc-modify-drafts-shortcode-user-id.php
+++ b/gp-advanced-save-and-continue/gpasc-modify-drafts-shortcode-user-id.php
@@ -1,0 +1,15 @@
+<?php
+
+/**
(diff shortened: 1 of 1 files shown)


veryspry:
Given that the `user_id` arg can actually just be passed in with the shortcode string, I will likely close this.

See this plugin PR for more context: https://github.com/gravitywiz/gp-advanced-save-and-continue/pull/5

PR #627, opened 2023-02-17
`gpasc-gf-salesforce-compatibility.php`: Refactored file for better clarity and added link to 3rd party script.

veryspry:
## Summary

Dave pointed out that the naming here was a little confusing so this aims to fix that!


spivurno (approved):
Sexy!

PR #628, opened 2023-02-17
`gpnf-triggered-population.php`: Fixed an issue with trigger creating duplicate child entries.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2156864305/44297?folderId=3808239

## Summary

Prevent the creation of duplicate child entries on the trigger.

Before:
https://www.loom.com/share/e991e8d37e874ae68479c2164d7a0f4e

After:
https://www.loom.com/share/061e642c7c854fc49faf6792da315f6e

### DIFF
--- a/gp-nested-forms/gpnf-triggered-population.php
+++ b/gp-nested-forms/gpnf-triggered-population.php
@@ -78,7 +78,7 @@ class GPNF_Triggered_Population {
 
 						$( '#field_{0}_{1}'.format( self.formId, self.triggerFieldId ) ).find( 'input' ).on( 'change', function() {
 							// disable the field while entry creation is processing to avoid duplicate entries
-							$( $( '#field_{0}_{1}'.format( self.formId, self.triggerFieldId ) ).attr( 'disabled', true ) );
+							$( '#field_{0}_{1}'.format( self.formId, self.triggerFieldId ) ).prop( 'disabled', true );
 							var input = $( this );
 							var value = input.val();
 							var checked = input[0].checked;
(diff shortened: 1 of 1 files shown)


gp-nested-forms/gpnf-triggered-population.php
```diff
@@ -119,6 +121,8 @@ public function output_script() {
 								window.gpnf_triggered_population_entry = response.data
 								GPNestedForms.loadEntry( response.data );
 							}
+							// re-enable the field after entry creation is processed
+							$( $( '#field_{0}_{1}'.format( self.formId, self.triggerFieldId ) ).attr( 'disabled', false ) );
```
claygriffiths, line 125: You're double-wrapping the `$()` here (and above), which isn't needed.

I can also see `$( '#field_{0}_{1}'.format( self.formId, self.triggerFieldId )` being moved to a `var` being helpful at this point 😄

saifsultanc:
@claygriffiths 

Same customer came back with an edge case wherein the trigger field is also the data to be copied:
https://secure.helpscout.net/conversation/2212721089/46731?folderId=7098280

Since, we serialize() the form data when the trigger value is hidden, it isn't copied. This version# 2 update fixes. Here is a summary from the starting to the last solution to the current update (~100 seconds):

https://www.loom.com/share/ec425f4d461b4f8db7032728eb40f4d3

gp-nested-forms/gpnf-triggered-population.php
```diff
@@ -103,13 +105,38 @@ public function output_script() {
...
+										data[name] = value;
+									}
+								} else {
+									// for other input fields, include the value as-is
+									data[name] = value;
+								}
+							});
+
+							return $.param(data);
+						};
```
claygriffiths, line 134: I'm not 100% opposed to having this here, but what if you move the `$form.serialize()` call before the input gets disabled?

saifsultanc:
Fixed with another PR

PR #629, opened 2023-02-17
`gpasc-enable-or-disable-on-all-forms.php`: Added new snippet to allow GPASC to be easily turned on/off on all forms with a single command.

veryspry:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2156846467/44296?folderId=6987275 

## Summary

Allows you to leverage WP CLI to easily turn GPASC on/off with a single command. See the new snippet header for example usage.

### DIFF
--- a/gp-advanced-save-and-continue/gpasc-enable-or-disable-on-all-forms.php
+++ b/gp-advanced-save-and-continue/gpasc-enable-or-disable-on-all-forms.php
@@ -1,0 +1,73 @@
+<?php
+
+/**
+ * Gravity Perks // File Renamer // Enable or Disable on All Forms
+ * https://gravitywiz.com/documentation/gravity-forms-file-renamer/
+ *
+ * Allows you to easily enable or disable GPASC on all forms with a single command. Note, that this requires you
+ * to use WP CLI.
+ *
+ * Usage:
+ *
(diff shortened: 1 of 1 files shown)


claygriffiths (commented):
This is awesome, nice work!

I'd put a...

```php
    if ( ! defined( 'ABSPATH' ) ) {
        exit; // Exit if accessed directly
    }
```

...at the top of the file in case the file publicly accessible.

Also, just a heads-up, there are some pretty rad APIs for creating custom commands with WP-CLI: https://make.wordpress.org/cli/handbook/guides/commands-cookbook/. Not sure it's worth porting to that but food for thought if you write another one of these!

veryspry:
> @veryspry On second thought, is there any reason why a runtime filter couldn't be used instead to disable it for all (or specific forms)? That way, it doesn't modify the actual data?

Not a bad idea! That said, the customer isn't actually going to use this so I'm not going to invest the time into that right now.

@spivurno should we add this to our list for the stable release?

spivurno:
@veryspry My gut is to wait until another customer requests it. 🙏

veryspry:
Closing as we (or other users) do not actually need this at the moment.

PR #630, opened 2023-02-20
`gpnf-close-modal-on-esc-or-outside-click.js`: Added snippet to close modal when ESC key is pressed, or when mouse clicked outside modal area.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2158727063/44397?folderId=3808239

## Summary

Close the modal when either escape key is pressed, or any click is done outside the modal area.

Here is how it works (~15 seconds):
https://www.loom.com/share/ab0e3a42d9a64652bd485840c218d459

### DIFF
--- a/gp-nested-forms/gpnf-close-modal-on-esc-or-outside-click.js
+++ b/gp-nested-forms/gpnf-close-modal-on-esc-or-outside-click.js
@@ -6,20 +6,16 @@
  *
  * 1. Install this snippet with our free Custom JavaScript plugin.
  *    https://gravitywiz.com/gravity-forms-custom-javascript/
+ * 2. Also add "Load Init Scripts Early" snippet to your theme's functions.php.
+ *    https://github.com/gravitywiz/snippet-library/blob/master/experimental/gfjs-early-init-scripts.php
  */
-// Replace 1 with Nested Form Field ID
-var gpnf = window['GPNestedForms_GFFORMID_1'];
-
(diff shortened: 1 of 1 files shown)


claygriffiths (changes requested):
@saifsultanc Nice work! I respect the brute-force approach.

That said, this can be drastically simplified using the `gpnf_modal_args` JS filter hook (it's undocumented, unfortunately) and then modifying the [Tingle.js](https://tingle.robinparisi.com/) args to accept `overlay` and `escape` in addition to `button` for `closeMethods`.

e.g. (this is untested)

```js
window.gform.addFilter( 'gpnf_modal_args', function( args, formId, fieldId, gpnf ) {
    // Only run for parent form ID 1. Remove this if the behavior is desired for all forms.
    if ( formId != 1 ) {
        return args;
    }

    args.closeMethods = ['overlay', 'button', 'escape'];

    return args;
} );
```

saifsultanc:
> @saifsultanc 

Gotcha! Always something new to learn :)

Though, this filter doesn't seem to be working at the moment; but I will dig it further. Thanks!

claygriffiths:
@saifsultanc Hmm, I wonder if it has to do with init order. Does this snippet help? 

https://github.com/gravitywiz/snippet-library/blob/master/experimental/gfjs-early-init-scripts.php

claygriffiths (approved):

PR #634, opened 2023-02-22
`gpasc-use-submission-values-for-draft-names.php`: Added example snippet to use draft submission values as draft names.

veryspry:
## Context

⛑️ Ticket(s):
* https://secure.helpscout.net/conversation/2163412284/44602/
* https://secure.helpscout.net/conversation/2164442912/44658/

## Summary

Uses draft submission values for draft display names.

### DIFF
--- a/gp-advanced-save-and-continue/gpasc-use-submission-values-for-draft-names.php
+++ b/gp-advanced-save-and-continue/gpasc-use-submission-values-for-draft-names.php
@@ -14,9 +14,12 @@ add_filter( 'gpasc_draft_display_name_123', function( $display_name, $form_id, $
 	$submission_data   = GFFormsModel::get_draft_submission_values( $resume_token_data['token'] );
 	$submission_values = json_decode( $submission_data['submission'], true );
 	$submission_value  = rgar( $submission_values['submitted_values'], $field_id );
+	// If the field is a single input field like a text field.
 	if ( ! is_array( $submission_value ) ) {
 		empty( $submission_value ) ? $display_name : $display_name .= ' - ' . $submission_value;
 	} else {
+		// If it's a multi-input field like a Name field.
+		// This gets the first name like so field_id.3 and the last_name like so field_id.6.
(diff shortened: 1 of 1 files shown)


gp-advanced-save-and-continue/gpasc-use-submission-values-for-draft-names.php
```diff
@@ -0,0 +1,24 @@
...
+ * Use draft submission values to populate the draft name.
+ */
+// Update "123" to your form ID.
+add_filter( 'gpasc_draft_display_name_123', function( $display_name, $form_id, $resume_token_data ) {
+	// Update "4" to the ID of the field to be used to set the draft name.
+	$field_id          = 4;
+	$submission_data   = GFFormsModel::get_draft_submission_values( $resume_token_data['token'] );
+	$submission_values = json_decode( $submission_data['submission'], true );
+	$submission_value  = rgar( $submission_values['submitted_values'], $field_id );
+	if ( ! is_array( $submission_value ) ) {
```
veryspry, line 17: Can you add a comment here to let users know that this condition would be triggered if using on a single line text field? And then similarly, add a comment to the `else` block with an explanation of how you are grabbing values for the first/last name fields?

These are useful because it sometimes helps others feel confident modifying the scripts themselves for their exact use case.

veryspry:
@sbassah Nice work!

Can you send this over to the customer to see if it will work out for them? I made a Loom showing how we usually send the code over: https://www.loom.com/share/e68e196858154b7ea815420cd04cb1f6

gp-advanced-save-and-continue/gpasc-use-submission-values-for-draft-names.php
veryspry: Also, this is not needed to send to the customer to test!

sbassah: Well Noted. I will send it over to the customer to test.

veryspry:
@sbassah thanks again for the help finishing this - it looks great! :raised_hands:

claygriffiths (approved):

PR #636, opened 2023-02-27
`gpafp-replace-space-separator.php`: Added a new snippet to replace the space separator.

sbassah:
## Context

This snippet replaces the space separator when using the phone[nationalNumberFormatted] modifier.

Ticket: https://secure.helpscout.net/conversation/2164492172/44662?folderId=3775888


spivurno (approved):
Good work! 

I went ahead and removed the plugin header since this is a snippet that is unlikely to be installed as a plugin and having a smaller header makes the snippet easier to read and modify.

PR #639, opened 2023-03-08
`gpml-ajax-upload.php`: Fixed an issue where the temporary file would be removed too soon which could cause validation errors and prevent post images from being set correctly.

claygriffiths:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2155824919/44253?folderId=1530128

## Summary

Fixed an issue where the temporary file would be removed too soon, which could cause validation errors and prevent post images from being set correctly.

### DIFF
--- a/gp-easy-passthrough/gpep-edit-entry.php
+++ b/gp-easy-passthrough/gpep-edit-entry.php
@@ -120,7 +120,7 @@ class GPEP_Edit_Entry {
 
 		$entry_ids = array();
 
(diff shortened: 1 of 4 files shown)


claygriffiths:
This PR shouldn't be needed after a related issue is resolved in GF core.

claygriffiths:
Closing this PR since https://github.com/gravityforms/gravityforms/pull/2572 has been merged

PR #640, opened 2023-03-09
`gw-populate-date.php`: Fixed an issue with same source and target fields not working with different modifiers.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2177018899/45180?folderId=3808239

## Summary

The script slug generated only took into account the source field id and target field id. Thus, when we had a scenario wherein the same source and target fields were aimed at, but with different modifier fields only the first one programmatically would work. This adds the modifier input id to the script slug thus resolving the issue.

Here is how it works now:

https://www.loom.com/share/cde9a6c3f69d4be69238a1fc25c586dd

### DIFF
--- a/gravity-forms/gw-populate-date.php
+++ b/gravity-forms/gw-populate-date.php
@@ -1864,7 +1864,7 @@ class GW_Populate_Date {
 		);
 
 		$script = 'new GWPopulateDate( ' . json_encode( $args ) . ' );';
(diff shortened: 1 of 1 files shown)


claygriffiths (approved):
LGTM!

One small detail, I'd put in an `array_filter` on the array since the inputId might be null.

Here's what I mean:

```php
// string(6) "a_b_c_"
var_dump( implode( '_', array( 'a', 'b', 'c', null ) ) );

// string(5) "a_b_c"
var_dump( implode( '_', array_filter( array( 'a', 'b', 'c', null ) ) ) );
```

PR #645, opened 2023-03-27
`gw-manual-entries.php`: Fixed an issue with Manual Entries button diplaying on Stripe Sales View.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2182975851/45466?folderId=7098280

## Summary

 The "Add New Entry" button shows on the Stripe Sales screen when the snippet is enabled. This update fixes that.

**_BEFORE:_**
Screenshot 2023-03-27 at 4 04 05 PM

**_AFTER:_**
Screenshot 2023-03-27 at 4 04 21 PM

### DIFF
--- a/gravity-forms/gw-manual-entries.php
+++ b/gravity-forms/gw-manual-entries.php
@@ -42,6 +42,11 @@ class GW_Manual_Entries {
 			return;
 		}
 
+		// do not display the Add New Entry button for Stripe Sales Screen
+		if ( rgget( 'view' ) == 'gf_results_gravityformsstripe' ) {
+			return;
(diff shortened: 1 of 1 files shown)


saifsultanc:
@claygriffiths The one thing that we could probably look at it is displaying it for all 'views', so only display if rgget( 'view' ) is empty. But I am not sure if it would conflict with other scenarios ?

claygriffiths (approved):

claygriffiths:
@saifsultanc I think the current approach is fine. We can circle back if we get more requests.

PR #646, opened 2023-03-28
`gpnf-sync-parent-child-entry-payment-status.php`: Fixed a fatal error when editing entry.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2187168189/45727/

## Summary

Fixed a fatal error that is caused when we were editing an entry.


claygriffiths (changes requested):
While this works, I think the root issue is `gform_after_update_entry` passing `form, $lead['id'], $original_entry` and this function accepting `$entry_id`

saifsultanc:
Addressed in other PRs.

PR #648, opened 2023-04-05
`gw-save-and-continue-auto-load.php`: Fixed an issue with Save & Continue Auto Load snippet activating on forms it should not.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2198642933/46143/

## Summary

There isn't an explicit “Save & Continue” feature for the form in process, which may cause multi-page forms to “remember” values from previous partial submissions when they shouldn’t. This update adds that check, keeping the functionality same.

### DIFF
--- a/gravity-forms/gw-save-and-continue-auto-load.php
+++ b/gravity-forms/gw-save-and-continue-auto-load.php
@@ -166,7 +166,8 @@ class GW_Save_Continue_Auto_Load {
 	public function is_applicable_form( $form ) {
 		$form_id = isset( $form['id'] ) ? $form['id'] : $form;
 
-		if ( $this->is_editing_entry( $form_id ) ) {
+		$form = GFAPI::get_form( $form_id );
+		if ( $this->is_editing_entry( $form_id ) || ! $form['save']['enabled'] ) {
 			return false;
 		}
 


gravity-forms/gw-save-and-continue-auto-load.php
```diff
@@ -166,7 +166,8 @@ public function is_editing_entry( $form_id ) {
 	public function is_applicable_form( $form ) {
 		$form_id = isset( $form['id'] ) ? $form['id'] : $form;
 
-		if ( $this->is_editing_entry( $form_id ) ) {
+		$form = GFAPI::get_form( $form_id );
+		if ( $this->is_editing_entry( $form_id ) || ! $form['save']['enabled'] ) {
 			return false;
 		}
```
claygriffiths, line 172: `rgars( $form, 'save/enabled' )` is likely the safer option here. It's what I see used in GF core.

I'd also group the variables like so:

```php
$form_id = isset( $form['id'] ) ? $form['id'] : $form;
$form    = GFAPI::get_form( $form_id );

if ( $this->is_editing_entry( $form_id ) || ! rgars( $form, 'save/enabled' ) ) {
	return false;
}
```

claygriffiths (approved):
Good stuff! Giving this a preemptive approval pending my suggestions.

PR #649, opened 2023-04-06
`gf-openai-post-save-result-to-field.php`: Added snippet to populate fields with Open AI generated values.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2190755593/45844?folderId=7098280

## Summary

Add support to populate fields with Open AI generated values.

### DIFF
--- a/gravityforms-openai/gf-openai-post-save-result-to-field.php
+++ b/gravityforms-openai/gf-openai-post-save-result-to-field.php
@@ -1,0 +1,16 @@
+<?php
+/**
+ * Gravity Perks // GravityForms-OpenAI // Add support to populate fields with Open AI generated values.
+ *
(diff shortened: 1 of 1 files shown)


saifsultanc:
> Not a proper review but just high level feedback that I would provide clarification in the snippet description on the scenario in which this would be useful. Without clarification, it'll be a bit confusing since the plugin does provide an option for saving results to a field in the feed settings.

Makes sense @spivurno ! I have gone ahead and make a Loom for it. Do you think that would suffice?

saifsultanc:
> Love the video! But I still think a brief description in the header comment would be super helpful. Most users won't be inclined to watch the video if they don't understand why they'd need the snippet.

Added @spivurno

PR #656, opened 2023-04-27
`gw-update-user-password-after-submission.php`: Added new snippet for…

coulterpeterson:
… GF that validates two passwords are the same, and then updates the password for the provided email.

Source: https://gist.github.com/coulterpeterson/2d8cfbae77fd93bdd0744d4db4545daf

## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2223492574/47491/

## Summary

Pushing the initial version of this snippet. It updates the password for the user ID associated with the inputted email, allowing a custom password reset form.

### DIFF
--- a/gravity-forms/gw-update-user-password-after-submission.php
+++ b/gravity-forms/gw-update-user-password-after-submission.php
@@ -7,8 +7,15 @@
  * 
  */
 
-// Update "123" to your form ID, "4" to your "Password" field ID, and "5" to your "New Password" field ID.
+// Update "123" to your form ID, "4" to your "Password" field ID, and "5" to your "Confirm Password" field ID.
+// If you're not using a "Confirm Password" field, change "$double_password_confirmation" from "true" to "false".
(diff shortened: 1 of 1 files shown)


spivurno:
Untested but thoughts on using an Update User feed (via the User Registration add-on) and mapping a Password field to the Password setting in the feed? Would that work for this customer's use case?

coulterpeterson:
@spivurno just played around with it and I'm pretty sure GF User Registration only handles updating the currently logged-in user. The customer's use case was a hidden form where any WordPress user's password could be changed.

spivurno:
Ah, that makes sense. 

I am a little concerned about the security implications here. If the user misconfigures their form by not requiring a login or populating administrative users, they could unintentionally give users the ability to set an administrator's password and gain access to that account.

My gut is rather than investing additional effort hardening this up as a general purpose snippet, you could keep it as a personal gist and we can revisit based on demand. Thoughts?

coulterpeterson:
Makes sense to me! 😃

PR #664, opened 2023-05-12
`gpml-apc-attach-files-to-post.php`: Added a snippet to attach files to posts created with Gravity Forms APC and ACF.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2236815395/48148/

## Summary

Attach Files uploaded via GPML to Posts created with GFAPC + ACF.

https://www.loom.com/share/7087869f092c4b9ea5c18e92d7ca1469

https://www.loom.com/share/f2480807382f4fd2a622db0f89e65351

### DIFF
--- a/gp-media-library/gpml-apc-attach-files-to-post.php
+++ b/gp-media-library/gpml-apc-attach-files-to-post.php
@@ -1,7 +1,7 @@
 <?php
 /**
  * Gravity Wiz // Gravity Forms // Attach Files uploaded via GPML to Posts created with GFAPC + ACF
- * 
+ *
  * Attach Files uploaded via GPML to Posts created with GFAPC + ACF
  *
  * Instructions:
@@ -21,7 +21,6 @@ add_action( 'gform_advancedpostcreation_post_after_creation', function( $post_id
 					'post_parent' => $post_id,
 				) );
 			}
-			set_post_thumbnail( $post_id, $file_id );
 		}
 	}
 }, 10, 4 );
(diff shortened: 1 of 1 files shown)


gp-media-library/gpml-apc-attach-files-to-post.php
```diff
@@ -0,0 +1,27 @@
...
+
+		if ( $is_media_library ) {
+			$file_ids = gp_media_library()->get_file_ids( $entry['id'], $field->id );
+			foreach ( $file_ids as $file_id ) {
+				wp_update_post( array(
+					'ID'          => $file_id,
+					'post_parent' => $post_id,
+				) );
+			}
+			set_post_thumbnail( $post_id, $file_id );
```
claygriffiths, line 24: This would use the last `$file_id` as the featured image.

I think for `set_post_thumbnail`, we'd need to find the field that it mapped to it a la https://github.com/gravitywiz/gp-media-library/blob/master/class-gp-media-library.php#L672

spivurno: There is no need to set the post thumbnail. The user only wants to attach the image to the post which is handled by the function above.

saifsultanc: @claygriffiths @spivurno darn! I added it by mistake. I know we don't need it, didn't use it for testing anyways. Will update today

claygriffiths (approved):

PR #665, opened 2023-05-16
`gpnf-auto-attach-child-fields.php`: Improved compatibility with File Upload fields using GP Media Library.

claygriffiths:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2230815348/47796?folderId=1530128

## Summary

Improve compatibility with File Upload fields using GP Media Library.


spivurno (approved):
LGTM!

What value was being returned before this?

claygriffiths:
@spivurno The issue is that `handle_attachments` in GPML is looking at the parent form entry for file IDs and does things like looking for file upload fields. This doesn't work as we're essentially injecting these attachments into the parent form.

PR #670, opened 2023-05-31
Create gpld-allow-selecting-days-from-other-months.js

coulterpeterson:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2258552313/49352/

## Summary

Allow days from other months to be selectable in the calendar month view, without having to click next/previous month.


spivurno (changes requested):
My man! 🙌

Since we're going to be rolling this into the Snippet Library, could see if this is possible with the [gform_datepicker_options_pre_init](https://docs.gravityforms.com/gform_datepicker_options_pre_init/) filter? I _think_ that approach might also make this work with non-inline datepickers too.

PR #671, opened 2023-06-05
`gpnf-child-entry-id-modifier.php`: Fixed an issue with character encoding.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2261594766/49479?folderId=3808239

## Summary

Fixed an issue with character encoding. Using `DOMDocument::load` implies the text will be regarded in ISO-8859-1, which is the HTTP/1.1 standard character set. This results in UTF-8 strings being misinterpreted. 

To resolve this, we prepend an XML encoding declaration to the text

**_BEFORE:_**
Screenshot 2023-06-05 at 3 15 36 PM

**_AFTER:_**
Screenshot 2023-06-05 at 3 15 17 PM

### DIFF
--- a/gp-nested-forms/gpnf-child-entry-id-modifier.php
+++ b/gp-nested-forms/gpnf-child-entry-id-modifier.php
@@ -19,7 +19,7 @@ add_filter( 'gpnf_all_entries_nested_entry_markup', function( $markup, $field, $
 	}
 
 	$dom = new DOMDocument();
-	$dom->loadHTML( $markup );
+	$dom->loadHTML( '<?xml encoding="utf-8" ?>' . $markup );
 
 	$first_row  = $dom->getElementsByTagName( 'tr' )->item( 1 );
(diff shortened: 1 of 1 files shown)


gp-nested-forms/gpnf-child-entry-id-modifier.php
```diff
@@ -19,7 +19,7 @@
 	}
 
 	$dom = new DOMDocument();
-	$dom->loadHTML( $markup );
+	$dom->loadHTML( '<?xml encoding="utf-8" ?>' . $markup );
```
claygriffiths, line 22: This is an XML declaration which might cause some unforeseen issues. Before pressing on with this, can you give this a try?

Change `$dom = new DOMDocument();` to `$dom = new DomDocument('1.0', 'UTF-8');`

Either way, nice work finding the issue!

PR #672, opened 2023-06-14
`gw-edit-product-and-payment-details.php`: Fixed issue with GPECF where "subtotal" fields would not be saved correctly after editing a product field on an entry.

veryspry:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2267707761/49837/

## Summary

This fixes an issue where `GW_Edit_Products` was not updating the "subtotal" fields when entries were being edited.

The result is that, after editing an entry, the subtotal columns in the entry table view would be blank like this:

image

However, the entry table view should show a value in each subtotal cell like this:

Screenshot 2023-06-14 at 2 53 21 PM

### DIFF
--- a/gravity-forms/gw-edit-product-and-payment-details.php
+++ b/gravity-forms/gw-edit-product-and-payment-details.php
@@ -106,8 +106,6 @@ class GW_Edit_Products {
 		}
 
 		if ( $has_product_field ) {
-			// calculate the total once product fields have been restored to their original types
-			$total = GFCommon::get_order_total( $form, $entry );
 			foreach ( $form['fields'] as &$field ) {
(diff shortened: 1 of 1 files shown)


spivurno (approved):
Looks good. 🤘

Would have thought there'd be some sort of helper function for this in GP_Ecommerce_Fields but doesn't look like it. 👎

veryspry:
Looks like there is _one_ more case where this needs to be updated (see [ticket](https://secure.helpscout.net/conversation/2267707761/49837/) for details)

veryspry:
> Would have thought there'd be some sort of helper function for this in GP_Ecommerce_Fields but doesn't look like it. 👎

I looked and thought the same thing 😞

PR #674, opened 2023-06-15
Update GW_Coupons_Exclude_Products to support GF Coupons 3.1 and add new functionality

FPCSJames:
Replaces #673 - patch branch renamed.

## Context

The Gravity Forms team added formId as a parameter of the gform_coupons_discount_amount JS filter in GF Coupons 3.0.1. This snippet no longer works as of that version and above - arguments.callee.caller.caller.caller.caller causes a fatal error.

## Summary

This PR contains some extra goodies atop the GFC 3.1 support:

- Multi-form support: previously, to use this snippet with multiple forms, it would have to be instantiated multiple times, and in doing so the excluded total could be wrong in some rare edge cases. Now, it can be instantiated once (more efficient) with the exclude_fields_by_form param set.
- Optionally skip exclusion for 100% off coupons: it's likely (but not guaranteed) that an operator intends for a 100% off coupon to make products on the form fully free, which means even total from the excluded fields needs to be discounted. _Example:_ a calculation product is in place to pass along Stripe fees with the formula `((({subtotal} - {discounts}) + 0.30) / 0.965) - ({subtotal} - {discounts})`, and this product is excluded from coupons. This works correctly until a 100% off coupon is used - if the exclusions aren't skipped, the final total will be $0.30, not $0. Handling this scenario would otherwise require an extra field with `{subtotal}` in it, to enable conditional logic on the fee field, which is not ideal.
- Small fixes.


spivurno:
@FPCSJames Thanks for the PR! You seem to be a very meticulous person. I like it! 

I'll review this next week and get back to you. 🙏

spivurno:
Apologies for the delay getting to this, James. Still on my radar. 🙏

spivurno:
Well, @FPCSJames, this completely fell off of my radar. But... merged. 😅

PR #677, opened 2023-06-19
`gppa-object-type-json-api.php`: Fixed an issue with the snippet not working with GPPA 2.0+.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2273072569/50303/
https://secure.helpscout.net/conversation/2275362111/50406?folderId=3808239

## Summary

The snippet fails to work with GPPA 2.0, because GPPA 2.0 also stores the 'limit' value in the same array. A simple rule, would be as follows:

```
array(
    0 => array(
        0 => array(
            'property' => 'name',
            'operator' => 'contains',
            'value' => 'faa'
        )
    ),
    'limit' => 750
);
```

This update adds a conditional logic to skip processing for non-array values.


claygriffiths (approved):
LGTM!

PR #681, opened 2023-06-28
Update gpnf-attach-child-entry-programatically.php

pimteam:
Should be add_filter instead of add_action

## Context


spivurno (approved):
'tis true! 

Thanks for the PR!

PR #685, opened 2023-07-09
`gravity-forms/gw-multi-file-merge-tag.php:` Added support for Image Hopper fields.

jakejackson1:
## Context

Had a Gravity Wiz customer request this functionality via Image Hopper support.

## Summary

Allow users to change the merge tag template for both File Upload fields and Image Hopper fields (uses the exact same underlying storage mechanism as a Multi File Upload field). 

This change allows other fields to be easily added in the future, should support need to be added.


spivurno (approved):
LGTM!

PR #689, opened 2023-07-18
`gw-prevent-duplicate-drop-down-selections.js`: Fixed an issue where options disabled by GP Inventory could get re-enabled even if the choice inventory was at zero.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2292978921/51848?folderId=3808239

## Summary

When using [this snippet](https://github.com/gravitywiz/snippet-library/blob/master/experimental/gw-prevent-duplicate-drop-down-selections.js) on fields that have inventory set by GP Inventory, the snippet 're-enables' options that have been disabled by GP Inventory because they are out of stock, allowing a user to select options that they shouldn't be able to.

Here is quick loom of resolving this issue (~100 seconds):
https://www.loom.com/share/b83b90597ba2461bb244b5e140f69663

### DIFF
--- a/experimental/gw-prevent-duplicate-drop-down-selections.js
+++ b/experimental/gw-prevent-duplicate-drop-down-selections.js
@@ -23,6 +23,10 @@ jQuery(".gfield_ddselect").change(function () {
 	jQuery(".gfield_ddselect").each(function (a) {
 		var parent_el = jQuery(this);
 		jQuery("option", jQuery(this)).each(function (b) {
+			// With GP Inventory, any Inventory Exhausted choice should remain disabled.
+			if ($(this).attr('class') != 'gpi-disabled') {
+				jQuery(this).prop("disabled", false);
+			}
 			if (b != 0) {
 				for (c = 0; c < select_num; c++) {
 					if (a != c) {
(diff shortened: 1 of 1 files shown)


experimental/gw-prevent-duplicate-drop-down-selections.js
```diff
@@ -23,7 +23,6 @@ jQuery(".gfield_ddselect").change(function () {
 	jQuery(".gfield_ddselect").each(function (a) {
 		var parent_el = jQuery(this);
 		jQuery("option", jQuery(this)).each(function (b) {
-			jQuery(this).prop("disabled", false);
```
claygriffiths, line 26: I think removing this would likely break some existing functionality.

I'd just check for the `gpi-disabled` class on `this` and if it does have the class, then do not remove the disabled prop.

spivurno: @saifsultanc Just confirming Clay's suspicion that this breaks the core functionality of this snippet: https://www.loom.com/share/5fafb1774df342049639cb5e76e114d5

experimental/gw-prevent-duplicate-drop-down-selections.js
```diff
@@ -23,7 +23,10 @@ jQuery(".gfield_ddselect").change(function () {
...
+			// With GP Inventory, any Inventory Exhausted choice should remain disabled.
+			if ($(this).attr('class') != 'gpi-disabled') {
```
claygriffiths, line 27:
```suggestion
			if (! $(this).hasClass('gpi-disabled')) {
```

claygriffiths (approved):
Feel free to squash and merge and testing and implementing this last suggestion.

With `hasClass()`, it can support there being multiple classes for the option vs. a simple string comparison on the class attribute.

PR #690, opened 2023-07-18
`gpecf-set-discount-amount-by-field-value.php`: Fixed an issue with the snippet not dynamically updating the discount value.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2299585841/52198?folderId=3808239

## Summary

When using the snippet, the Discount field's value isn't updating on the front end. It is only reflected after the form is submitted.

BEFORE:
https://www.loom.com/share/a9ecb44762c3400ba8321cc0de3e1759

AFTER:
https://www.loom.com/share/c1f2cc21fb0d47b59ba6d69571634bc3

### DIFF
--- a/gp-ecommerce-fields/gpecf-set-discount-amount-by-field-value.php
+++ b/gp-ecommerce-fields/gpecf-set-discount-amount-by-field-value.php
@@ -110,10 +110,10 @@ class GPECF_Discount_Amounts_By_Field_Value {
 						var amount = parseFloat(value, 10);
 						if ( !isNaN( amount ) ) {
 							var currency = new window.Currency(
-							window.gf_global.gf_currency_config
-						);
+								window.gf_global.gf_currency_config
+							);
 
-						var formattedAmount = currency.toMoney(amount);
+							var formattedAmount = currency.toMoney(amount);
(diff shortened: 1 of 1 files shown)


gp-ecommerce-fields/gpecf-set-discount-amount-by-field-value.php
```diff
@@ -107,6 +107,15 @@ public function output_script() {
...
 						self.$getInput( self.discountFieldId ).data( 'amount', value );
+						var amount = parseFloat(value, 10);
+						if ( !isNaN( amount ) ) {
+							var currency = new window.Currency(
+							window.gf_global.gf_currency_config
+						);
+
+						var formattedAmount = currency.toMoney(amount);
+							$( '.ginput_discount_' + self.formId + '_' + self.discountFieldId ).html( formattedAmount );
+						}
```
claygriffiths, line 118: The indention is a little wonky here. Can you make sure this is all formatted correctly?

I'm mildly surprised PHPCS isn't saying anything 😃

claygriffiths (commented):
The code looks good, but the indention isn't correct, which makes it confusing to read.

claygriffiths (approved):

PR #697, opened 2023-08-22
`gppa-encrypt-fileupload-field-links.php`: Added a snippet that encrypts file upload field links with GPPA.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2326751274/53373?folderId=3808239

## Summary

When using GPPA to get the URL of uploaded files in another form entry. GPPA returns the direct link to the file. This update adds the ability to get encrypted links with GPPA.

Snippet in action (~150 seconds) with **Before** and **After**:
https://www.loom.com/share/699da443af204e93b886c2d72e93d702

### DIFF
--- a/gp-populate-anything/gppa-encrypt-fileupload-field-links.php
+++ b/gp-populate-anything/gppa-encrypt-fileupload-field-links.php
@@ -1,0 +1,33 @@
+<?php
+/**
+ * Gravity Perks // Populate Anything // Encrypt File Upload Field Links
+ * https://gravitywiz.com/documentation/gravity-forms-populate-anything/
+ *
+ * Encrypt File Upload Links when populated with GP Populate Anything.
+ *
+ * Note: This will disable the default behavior of file upload field encryption on Gravity Forms.
+ * Instead, it will manually encrypt the file upload links directly on the Gravity Forms Entry.
+ */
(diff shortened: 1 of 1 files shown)


claygriffiths (commented):
I'm a bit weary of this approach as changing the actual saved value is likely going to cause issues.

Can we just use this logic to generate a secure URL using GPPA-specific filters?

gp-populate-anything/gppa-encrypt-fileupload-field-links.php
```diff
@@ -0,0 +1,33 @@
...
+/**
+ * Gravity Perks // Populate Anything // Encrypt File Upload Field Links
+ * https://gravitywiz.com/documentation/gravity-forms-populate-anything/
+ *
+ * Encrypt File Upload Links when populated with GP Populate Anything.
+ *
+ * Note: This will disable the default behavior of file upload field encryption on Gravity Forms.
+ * Instead, it will manually encrypt the file upload links directly on the Gravity Forms Entry.
+ */
+add_filter( 'gform_secure_file_download_location', '__return_false' );
```
claygriffiths, line 11: Disabling this across the board doesn't seem like a good idea.

saifsultanc:
customer never got back. It's been 18 months already.

PR #701, opened 2023-08-25
`gpfr-apply-rename-to-every-field.php`: Added snippet to apply GPFR to every field, no matter if there is an existing template value or not.

veryspry:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2330760345/53504?folderId=6987275 

## Summary

If using the `gpfr_filename_template` to automatically set a filename template for every field, it can become quite laborious to go through every single field and add a "dummy" template. This allows you to automatically apply GPFR to any field you want to on the fly.

🚨 Relies on the changes in this GPNF pull request: https://github.com/gravitywiz/gp-file-renamer/pull/13

### DIFF
--- a/gp-file-renamer/gpfr-apply-rename-to-every-field.php
+++ b/gp-file-renamer/gpfr-apply-rename-to-every-field.php
@@ -13,3 +13,20 @@
 add_filter( 'gpfr_is_applicable_field', function ( $should_apply_gpfr, $form, $field ) {
 	return true;
 }, 10, 3 );
+
+
+/**
+ * It is recommended to also register a filter callback for `gpfr_filename_template` to ensure that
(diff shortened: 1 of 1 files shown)


veryspry:
Just waiting on this change to get merged: https://github.com/gravitywiz/gp-file-renamer/pull/13

claygriffiths (approved):

veryspry:
> I'm curious if it would make more sense if this snippet also included a `gpfr_filename_template` example as well? What happens if only this snippet is installed?

Hmmmm good point. Will add something!

veryspry:
@claygriffiths I added the `gpfr_filename_template` example in 19a9a47789a2aaa22a6783cde8b5453f18b2cf6e. Good thing too since _not_ including that results in files that are renamed to empty strings when the field setting is empty 😅

PR #702, opened 2023-08-25
Fix compatibility issue with GPASC and GPNF.

veryspry:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2318327951/52962?folderId=6987275 

## Summary

Because this snippet loads the form markup via AJAX and then manually calls `gform_post_render` AFTER the request has succeeded. As such, some perk scripts need to wait to init until `gform_post_render` occurs.


veryspry:
Closing as we found a better solution for this!

PR #706, opened 2023-09-17
Fixes for count and currency parsing

damoncasale:
## Context

## Summary

This is a proposed fix for the aggregate functions snippet.  By default, the {count} merge tag requires that the gppa_query_all_value_objects filter return "true."  Because this snippet overrides that filter, it breaks the {count}
merge tag.  This fix resolves that issue.

This fix also resolves an issue with parsing currency values from fields that aren't "product" fields.  For instance, if we have a group of checkboxes with currency values in them, values with dollar signs or thousands separators won't be parsed properly.  Feel free to modify this portion of the fix to improve currency parsing.

### DIFF
--- a/gp-populate-anything/gppa-aggregate-functions.php
+++ b/gp-populate-anything/gppa-aggregate-functions.php
@@ -1,4 +1,4 @@
-<?php
+
 /**
  * Gravity Perks // GP Populate Anything // Add Support for Aggregate Functions
  * https://gravitywiz.com/documentation/gravity-forms-populate-anything/
@@ -22,6 +22,7 @@
  */
 class GPPA_Aggregate_Functions {
 
+	public static $default_count_regex = '/{count}/';
 	public static $sum_regex = '/{sum:(.+)}/';
 	public static $avg_regex = '/{avg:(.+)}/';
 	public static $min_regex = '/{min:(.+)}/';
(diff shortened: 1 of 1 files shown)


spivurno:
Thanks for the PR, @damoncasale. Reviewing now. 🙂

gp-populate-anything/gppa-aggregate-functions.php
```diff
@@ -163,6 +168,14 @@ public function get_object_value( $merge_tag, $object, $object_type ) {
...
 				$value = GFCommon::to_number( $value, $object->currency );
 			}
+		} else {
+			// Even if this isn't a product field, it might have a currency value in it.
+			// Handle it accordingly.
+			// TODO:  Handle other currency locales
+			$language = get_user_locale();
+			if (str_contains($language, "en")) {
+				$value = preg_replace("/[\$\,]/", "", $value);
+			}
```
spivurno, line 178: I'm not entirely comfortable with this logic. `get_user_locale()` can be different depending on the user which may work well for scenarios where the user is submitting entry values in their own currency but, typically, forms are configured to work in the locale of the site.

Without more real-world uses cases, my recommendation would be to ditch this `else` block and wrap the `$value` in a filter so you can apply this logic separately.

gp-populate-anything/gppa-aggregate-functions.php
```diff
@@ -1,4 +1,4 @@
-<?php
+
```
spivurno, line 1: We always include the opening <?php in our snippets.

gp-populate-anything/gppa-aggregate-functions.php
```diff
@@ -56,12 +56,17 @@ public function init() {
 
 	}
 
-	public function enable_query_all_value_objects( $query_all_value_objects, $field, $field_values, $object_type_instance, $filter_groups, $primary_property, $templates ) {
-		return $this->matches_any_merge_tag( rgar( $templates, 'value' ) );
+	public function enable_query_all_value_objects( $should_query_all, $field, $field_values, $object_type_instance, $filter_groups, $primary_property, $templates ) {
```
spivurno, line 59: Dig this variable name change. Definitely provides more clarity. 🤘

gp-populate-anything/gppa-aggregate-functions.php
```diff
@@ -56,12 +56,17 @@ public function init() {
...
+		return $should_query_all || $this->matches_any_merge_tag( rgar( $templates, 'value' ) );
```
spivurno, line 60: Busted! Good catch.

gp-populate-anything/gppa-aggregate-functions.php
```diff
@@ -56,12 +56,17 @@ public function init() {
...
 	}
 
+	// The count function normally has "query all value objects" enabled, but this function overrides that.
+	// We need to specify that count has to have "query all value objects" enabled to avoid breaking it.
```
spivurno, line 64: This comment is a bit too ambiguous to be useful for future us. My recommendation would be to ditch it.

spivurno (changes requested):

spivurno:
We

PR #709, opened 2023-09-26
`gw-submit-to-access.php`: Fixed a compatability issue between GW Submit to Access and Elementor Pro.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2365985616/54888/

## Summary

Elementor Pro is preventing GWSA from functioning. We need to leverage elementor's `elementor/frontend/the_content` hook in some way.


saifsultanc:
@claygriffiths I accidentally opened the PR instead of making it draft. Thoughts?

claygriffiths (approved):

claygriffiths:
Recommending the customer leverages `gwsa_requires_submission_redirect` for now. 

I explored looking into making it possible to use Elementor's CSS classes setting on element/widgets, but it's not trivial to filter the content of some of the more useful container elements such as columns.

PR #710, opened 2023-09-26
`gw-prioritize-next-over-save-and-continue.js`: Added a snippet to prevent the Save & Continue button from taking priority over Next buttons when pressing enter.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2371358392/55130?folderId=3808239

## Summary

In GPASC, there's an issue where pressing the Enter or Return key activates the "Save" action instead of the "Next" action. This is because the "Save and Continue" link is implemented using <input type="submit">, while the "Next page" button is implemented using <input type="button">. Overall this does seem the right approach.

To customize the user experience, this JavaScript snippet can be integrated to modify the behavior. This snippet enables users to effectively "skip" the default Enter key behavior associated with "Save and Continue" and instead allows them to progress through the form and navigate to the "Next" pages when they press Enter.

BEFORE:
https://www.loom.com/share/9e4bbabe16e346a28d70b96025d42306

AFTER:
https://www.loom.com/share/5922cba6fa5d43afa9d392e36301f725

### DIFF
--- a/gravity-forms/gw-prioritize-next-over-save-and-continue.js
+++ b/gravity-forms/gw-prioritize-next-over-save-and-continue.js
@@ -1,0 +1,11 @@
+/**
+ * Gravity Wiz // Gravity Forms // Prioritize Next Button Over Save and Continue When Pressing Enter
+ * https://gravitywiz.com/
+ *
+ * Instructions:
+ *    Install this snippet with our free Custom JavaScript plugin.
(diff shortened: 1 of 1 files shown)


claygriffiths:
From the testing I did, this issue appears to be present without GPASC activated, so it appears to be a core GF issue.

This snippet should be made a generic Gravity Forms snippet, moved into the `gravity-forms` folder, and prefixed with `gw` accordingly.

PR #712, opened 2023-09-27
`gpnf-populate-parent-entry-id.php`: Fixed issue where `gpnf_session_script_data_` filter callback was not returning the filtered value.

veryspry:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2372876723/55235/ 

## Summary

The `gpnf_session_script_data_` filter callback registered here was not returning a value which was breakin' some things.

### DIFF
--- a/gp-nested-forms/gpnf-populate-parent-entry-id.php
+++ b/gp-nested-forms/gpnf-populate-parent-entry-id.php
@@ -21,9 +21,10 @@ add_filter( 'gform_field_value_gpnf_parent_entry_id', function( $value, $field )
 		} else {
 			// The first time the parent form loads there will be no cookie. Create a hash and set it as the default
 			// hash for this cookie created for this form.
-			$value = $session->make_hashcode();
+			$value = $session->get_runtime_hashcode();
 			add_filter( "gpnf_session_script_data_{$field->formId}", function( $data ) use ( $value ) {
 				$data['hash'] = $value;
+				return $data;
 			} );
 		}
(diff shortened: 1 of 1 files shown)


gp-nested-forms/gpnf-populate-parent-entry-id.php
```diff
@@ -21,9 +21,10 @@
 		} else {
 			// The first time the parent form loads there will be no cookie. Create a hash and set it as the default
 			// hash for this cookie created for this form.
-			$value = $session->make_hashcode();
+			$value = $session->get_runtime_hashcode();
```
veryspry, line 24: I noticed that when we [set the default hashcode in GPNF](https://github.com/gravitywiz/gp-nested-forms/blob/53c13ffb0e10d66874d739e46f4b0a83d774a68a/includes/class-gpnf-session.php#L280), we call `get_runtime_hashcode()`. Does it make sense to do the same here? Or does it matter?

spivurno (approved):
Clean!

gp-nested-forms/gpnf-populate-parent-entry-id.php
spivurno: Yes! We added get_runtime_hashcode() later and it is the correct way to get the hashcode. 🙏

veryspry: Thank you for confirming!

PR #715, opened 2023-10-03
`gw-prevent-duplicate-selections.js`: Fixed an issue with placeholder value getting disabled.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2377401092/55467?folderId=3808239

## Summary

[This snippet](https://github.com/gravitywiz/snippet-library/blob/master/gravity-forms/gw-prevent-duplicate-selections.js) applies to Placeholder values, but it shouldn't.

Placeholder values should not be disabled.

**Both forms have the same placeholder values.**
Screenshot 2023-10-03 at 7 12 32 AM

**Before update** (on form load, second dropdown's Placeholder isn't selectable)
Screenshot 2023-10-03 at 7 16 56 AM

**After update** (on form load):
Screenshot 2023-10-03 at 7 12 57 AM

### DIFF
--- a/gravity-forms/gw-prevent-duplicate-selections.js
+++ b/gravity-forms/gw-prevent-duplicate-selections.js
@@ -47,7 +47,7 @@ function gwDisableDuplicates( $elem, $group, selected ) {
 	}
 
 	let value     = $elem.val();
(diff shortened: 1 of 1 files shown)


claygriffiths (approved):
Looks good!

PR #716, opened 2023-10-04
`gpcc-copy-label.js`: Fixed an issue with copying labels from checkbox selections.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2377837725/55492

## Summary

The snippet to copy the label of a choice-based field, when using Copy Cat, doesn't work on a Checkbox field. It only supports Dropdown and radio button fields. This update adds support for the checkbox field.

Screenshot 2023-10-04 at 8 32 10 PM

BEFORE:
https://www.loom.com/share/255f3dbe2e234d9694bf29ce3114a662

AFTER:
https://www.loom.com/share/731daa5781c1404489a290d83336c6ec

### DIFF
--- a/gp-copy-cat/gpcc-copy-label.js
+++ b/gp-copy-cat/gpcc-copy-label.js
@@ -19,7 +19,7 @@ gform.addFilter( 'gppc_copied_value', function( value, $elem, data ) {
 	} else if( $source.is( '.gfield_radio' ) ) {
 		value = $( '.gfield-choice-input:checked + label' ).text();
 	} else if ( $source.is( '.gfield_checkbox') ) {
-		$inputs= $source.find( 'input:checked' );
+		$inputs = $source.find( 'input:checked' );
 		var checkedLabels = [];
 		$inputs.each( function () {
 			var label = $(this).next('label').text();
--- a/gp-copy-cat/gpcc-copy-label.js
+++ b/gp-copy-cat/gpcc-copy-label.js
@@ -13,21 +13,19 @@
 * 2. Copy and paste the snippet into the editor of the Custom Javascript for Gravity Forms plugin.
 */
 gform.addFilter( 'gppc_copied_value', function( value, $elem, data ) {
-    $source = jQuery( '#input_' + data.sourceFormId + '_' + data.source );
-    if( $source.is( 'select' ) ) {
-        value = $source.find( 'option:selected' ).text();
-    } else if( $source.is( '.gfield_radio' ) ) {
-        value = $( '.gfield-choice-input:checked + label' ).text();
+	$source = jQuery( '#input_' + data.sourceFormId + '_' + data.source );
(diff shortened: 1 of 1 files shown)


gp-copy-cat/gpcc-copy-label.js
```diff
@@ -8,16 +8,26 @@
...
 gform.addFilter( 'gppc_copied_value', function( value, $elem, data ) {
     $source = jQuery( '#input_' + data.sourceFormId + '_' + data.source );
     if( $source.is( 'select' ) ) {
         value = $source.find( 'option:selected' ).text();
-    }  else if( $source.is( '.gfield_radio' ) ) {
+    } else if( $source.is( '.gfield_radio' ) ) {
         value = $( '.gfield-choice-input:checked + label' ).text();
+	} else if ( $source.is( '.gfield_checkbox') ) {
+		// Since multiple checkboxes maybe selected, make sure individual checkbox values do not have whitespaces.
+		values = value.split( ' ' );
```
claygriffiths, line 23: This feels a bit flimsy. Can you use something like `$source.find( 'input:checked' )` to get the checked inputs? 

Then, you can use a jQuery `.each()` on it and get the sibling label and put that into the `labelArray`.

gp-copy-cat/gpcc-copy-label.js
```diff
@@ -8,16 +8,26 @@
...
+		var labelArray = [];
+		values.forEach( function ( value ) {
+			$input = $( 'input[value="' + value + '"]' );
```
claygriffiths, line 26: This shouldn't matter based on my previous comment, but I wanted to call out that this selector would select across the whole document rather than the source field.

This would be better:

```
$input = $source.find( 'input[value="' + value + '"]' );
```

claygriffiths (changes requested):

gp-copy-cat/gpcc-copy-label.js
```diff
@@ -8,16 +8,24 @@
...
-        value = $source.find( 'option:selected' ).text();
-    }  else if( $source.is( '.gfield_radio' ) ) {
-        value = $( '.gfield-choice-input:checked + label' ).text();
+	$source = jQuery( '#input_' + data.sourceFormId + '_' + data.source );
+	if( $source.is( 'select' ) ) {
+		value = $source.find( 'option:selected' ).text();
+	} else if( $source.is( '.gfield_radio' ) ) {
+		value = $( '.gfield-choice-input:checked + label' ).text();
+	} else if ( $source.is( '.gfield_checkbox') ) {
+		$inputs= $source.find( 'input:checked' );
```
claygriffiths, line 22:
```suggestion
		$inputs = $source.find( 'input:checked' );
```

claygriffiths (approved):
LGTM other than one missing space.

claygriffiths (approved):

PR #717, opened 2023-10-04
Fixed issue where fields using the JSON API object without filters returned no results.

spivurno:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2377701857/55479/

## Summary

This PR addresses an issue introduced in https://github.com/gravitywiz/snippet-library/pull/677 where fields populating from the JSON API object without filters return no results. The issue was that we skipped that filter rather than indicating that all results matched that filter.

Worth noting this PR does not address the fact that the default limit of `501` is not honored by this snippet.


saifsultanc (approved):

spivurno:
Fixed by some other PR. 😅

PR #725, opened 2023-10-31
`gw-update-posts.php`: Fixed an issue with confirmation message not displaying.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2402316918/56523?folderId=3808239

## Summary

When the Update Posts snippet is active on this user's form, the confirmation message doesn't display after submitting the form.

We disable `wp_update_post` param `$fire_after_hooks` to ensure we do see the confirmation.

**Confirmation message setting on form:**
Screenshot 2023-10-31 at 2 10 06 PM

**BEFORE:**
Screenshot 2023-10-31 at 2 09 23 PM

**AFTER:**
Screenshot 2023-10-31 at 2 09 35 PM

### DIFF
--- a/gravity-forms/gw-update-posts.php
+++ b/gravity-forms/gw-update-posts.php
@@ -166,6 +166,7 @@ class GW_Update_Posts {
 
 		}
 
+		// ensure the fires after hooks is set to false, so that doesn't override some of the normal rendering - GF confirmation for instance.
 		wp_update_post( $post, false, false );
 
 	}
--- a/gravity-forms/gw-update-posts.php
+++ b/gravity-forms/gw-update-posts.php
@@ -166,7 +166,7 @@ class GW_Update_Posts {
 
 		}
 
-		wp_update_post( $post );
+		wp_update_post( $post, false, false );
 
 	}
 


spivurno:
@saifsultanc Why does this fix the issue?

claygriffiths:
@saifsultanc I am equally intrigued 😂. If this is the best path can you add an inline comment explaining why the params are there so we don't forget?

saifsultanc:
Sorry guys. Updated with the comment.

spivurno:
@saifsultanc I'm still looking for a better understanding of why this resolve the issue? The comment you've added describes what is failing but not why it was failing and how disabling the hooks resolves that.

claygriffiths:
Doing a bit of research, that parameter determines if `wp_after_insert_post()` is run, which runs the `wp_after_insert_post` action.

Interestingly, Gravity Forms does not use this hook and there are very few usages in WordPress that are added to this hook. 

The only other thing that I'm wondering if the issue could be related to is the `get_post()` call in https://developer.wordpress.org/reference/functions/wp_after_insert_post/#source which might be some sort of weird caching issue?

claygriffiths:
Admittedly, the logic around confirmations is quite complex, and I'm sure I could understand this with a 30 minute Xdebug session, but it's not worth the time if this resolves the issue for the customer.

spivurno:
@claygriffiths @saifsultanc I'm ok with moving forward with the existing solution. Let's just make it clear in the comment that this is a bit of voodoo so if there's an issue in the future we know we can challenge this parameter.

saifsultanc:
@claygriffiths @spivurno It was more of a hit and trial with different possible permutation to get that. I could debug to WHERE the issue was. Rest was to lookup documentation of wp_update_post and see what else that I can do without breaking current functionality and resolving the issue.

PR #726, opened 2023-11-01
`gpcld-enable-utc-timezone.js`: Removed due to deprecated filter.

scyt:
## Context

📓 Notion: https://www.notion.so/gravitywiz/Remove-snippet-from-library-fc80503b69e9448caa0511ee00ccc673?pvs=4

## Summary

Removing this snippet due to the filter being deprecated. It has been replaced with `gpcld_use_visitor_timezone`.


spivurno (changes requested):
Let's deprecate this instead of removing it.

https://www.notion.so/gravitywiz/Snippet-Library-Code-Organization-Formatting-and-Naming-d5bee101d9674dfc98d7ec007867de2b?pvs=4#1aed582f70e640e384a3a55562939926

spivurno (approved):
Approved with one tweak – can you add a link to the docs for that filter?

PR #728, opened 2023-11-09
`gppa-ignore-current-user-filter-for-admins.php`: Fixed an issue with snippet causing a fatal error.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2412200474/56932?folderId=3808239

## Summary

The snippet causes a fatal error when used on-site to remove the current user ID filter for admins.


gp-populate-anything/gppa-ignore-current-user-filter-for-admins.php
```diff
@@ -25,7 +25,7 @@
 		return $query_builder_args;
 	}
 
-	array_pop( $query_builder_args['where'][ $filter_group_index ] );
```
claygriffiths, line 28: Based on this usage, this is wiping out the entire `where` and effectively removing all filters from what I can see.

My guess is that the issue has to do with the filter using `special_value:current_user:ID` not being the last filter in the group.

You'll need to loop through `$query_builder_args['where'][ $filter_group_index ]` and find the filter that has `$filter['value'] === 'special_value:current_user:ID'` and remove it from the `$query_builder_args['where'][ $filter_group_index ]` array.

saifsultanc: Okay. Actually $filter_group_index does that job fine. What is happening specifically in this test scenario is that there is only 1 filter and that is using `special_value:current_user:ID`. So unsetting messes up the overall structure of the $query_builder_args which is what I've fixed in the next test and pushed it. @claygriffiths

gp-populate-anything/gppa-ignore-current-user-filter-for-admins.php
```diff
@@ -26,6 +26,9 @@
 	}
 
 	array_pop( $query_builder_args['where'][ $filter_group_index ] );
+	if ( count( $query_builder_args['where'][ $filter_group_index ] ) == 0 ) {
+		$query_builder_args['where'] = array();
+	}
```
claygriffiths, line 31: What if there are multiple groups?

For example, if `$filter_group_index` is `1` here, and we have another on `2`, this would clear that out as well.

PR #729, opened 2023-11-09
`gw-conditional-logic-enable-current-field.php`: Fixed an issue with conditional logic on current field snippet running on pages other than Form Editor.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2415455946/57066/

## Summary

Snippet breaking Conditional Logic on Notifications. The snippet taps the JavaScript hook `gform_is_conditional_logic_field` which runs on Notification page, but since that isn't a Form Editor page we need to add a logic to prevent it from firing on non form editor pages.

**BEFORE:**
Screenshot 2023-11-09 at 9 45 01 PM

**AFTER:**
Screenshot 2023-11-09 at 9 44 05 PM


claygriffiths (approved):
LGTM!

PR #732, opened 2023-11-24
`gw-all-fields-template.php`: Fixed an issue with the email field not processed as a text but a link.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2429986137/57820?folderId=3808239

## Summary

All Fields Template. Email fields are getting output as links. It should be getting output as an email text.

BEFORE:
Screenshot 2023-11-24 at 3 55 51 PM

AFTER:
Screenshot 2023-11-24 at 3 55 33 PM

### DIFF
--- a/gravity-forms/gw-all-fields-template.php
+++ b/gravity-forms/gw-all-fields-template.php
@@ -440,6 +440,7 @@ class GW_All_Fields_Template {
 					}
 
 					$raw_field_value = RGFormsModel::get_lead_field_value( $lead, $field );
+					$format          = $field->type === 'email' ? 'text' : $format;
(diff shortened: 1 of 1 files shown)


saifsultanc:
I have one concern with this update though, if we are using that same filter value `@{Nested Form:1:index[0],filter[3]}` on an HTML field (for instance), before this update it was getting output as follows:
Screenshot 2023-11-24 at 3 56 27 PM
Which looks right in the link form.

However, now it is output as follows (plain text)
Screenshot 2023-11-24 at 3 57 46 PM


Perhaps, in this case, it would make sense to have it as an HTML and not plain text. But then, flipside of this argument is that the user can very well add the `<a href` on their field pointing to the value returned by the filter `@{Nested Form:1:index[0],filter[3]}`

@claygriffiths

saifsultanc:
Addressed in another PR solution

PR #733, opened 2023-11-29
`gravity-forms/gw-prevent-duplicate-selections.js`: Added support for GPADVS enabled multi-select fields.

veryspry:
## Context

📓 Notion: https://www.notion.so/gravitywiz/GPADVS-1-1-Disabled-Choices-d910084a1ed24bd8947fdd6df242a2a9?pvs=4 

## Summary

This enabled support for this snippet on GPADVS enabled multi-select fields.


spivurno (approved):
Works flawlessly! 

I think we should convert this to a PHP snippet for delivery though. More coming in the card. That decision won't impact merging this though. Fire away!

PR #734, opened 2023-12-04
`gw-include-post-permalink.php`: Fixed an issue with Post ID not fetching with merge tags with APC addon updates.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2438396771/58350?folderId=3808239

## Summary

The Post permalink snippet doesn't work with the latest version of Gravity Forms Advanced Post Creation Add-On.

As a fix, this update adds a 'retry logic' to ensure we do fetch the post id. Changing the priority of the hooks `add_filter( 'gform_replace_merge_tags', array( $this, 'replace_merge_tag' ), 10, 3 );` wasn't giving a consistent resolution. The retry logic gave 100% results on my testing.

Video demo of the issue (Credits @sbassah):
 https://www.loom.com/share/dfc79bc1eda6424ab8bc7c4d457a6982?sid=7cf4436b-f7fe-434c-9a7b-07f3b4dbfa9a
 
 After this update, the merge tag always works.

### DIFF
--- a/gravity-forms/gw-include-post-permalink.php
+++ b/gravity-forms/gw-include-post-permalink.php
@@ -18,10 +18,17 @@
 class GWPostPermalink {
 
 	function __construct() {
-
+		add_filter( 'gform_is_feed_asynchronous', array( $this, 'disable_async' ), 10, 4 );
 		add_filter( 'gform_custom_merge_tags', array( $this, 'add_custom_merge_tag' ), 10, 4 );
 		add_filter( 'gform_replace_merge_tags', array( $this, 'replace_merge_tag' ), 10, 3 );
 
+	}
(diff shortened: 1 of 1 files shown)


spivurno:
Looks like GFAPC was updated to process its feeds asynchronously in 1.4. This would explain why your retry logic is working but also implies that it might be a bit of a brittle solution.

I wonder if the more bullet-proof solution might be to simply disable async feed processing? Could you do some research and see if it's possible to do so?

spivurno:
I think this is good for now! We can revisit if we see more demand or complaints about performance. 🤘

Only request is to add a comment to explain why we're disabling async.

PR #745, opened 2023-12-21
`gw-cache-buster.php`: Fixed issue where default `gform_post_render` event suppression incorrectly applied to all forms.

spivurno:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2438335097/58348

## Summary

Cache Buster incorrectly suppresses the default `gform_post_render` event for _all_ forms regardless of whether Cache Buster was enabled for that form or if that form was being rendered in a context that is applicable for Cache Buster.

Hilariously, I already tried to fix this once here (https://github.com/gravitywiz/snippet-library/commit/651990849fa8da6f9b5eacb3e0ad2945001aeeec) but I was clearly smoking crack at the time as it seems painfully obvious that was not the right place to move the filter. 😆


saifsultanc (approved):
Perfect!

PR #746, opened 2023-12-27
`gw-update-posts.php`: Fixed an issue with Update Posts updating post date time even when value was not set.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2461660694/59260?folderId=3808239

## Summary

The update post snippet also updates the Post Date to the current date even when the post date parameter is not set. This fix makes sure we check for the presence of the values provided in the argument before setting them, i.e. `$this->_args['post_date']['date']` and `$this->_args['post_date']['time']`.

### DIFF
--- a/gravity-forms/gw-update-posts.php
+++ b/gravity-forms/gw-update-posts.php
@@ -97,9 +97,13 @@ class GW_Update_Posts {
 		}
 
 		if ( $this->_args['post_date'] ) {
-			$new_date_time       = $this->get_post_date( $entry, $form );
-			$post->post_date     = $new_date_time;
-			$post->post_date_gmt = get_gmt_from_date( $new_date_time );
+			$new_date_time = $this->get_post_date( $entry, $form );
+			if ( $this->_args['post_date']['date'] ) {
+				$post->post_date = $new_date_time;
+			}
+			if ( $this->_args['post_date']['time'] ) {
+				$post->post_date_gmt = get_gmt_from_date( $new_date_time );
+			}
 		}
 
 		if ( $this->_args['featured_image'] && is_callable( 'gp_media_library' ) ) {
--- a/gravity-forms/gw-update-posts.php
+++ b/gravity-forms/gw-update-posts.php
@@ -11,6 +11,8 @@
  */
 class GW_Update_Posts {
 
+	protected $_args;
+
 	public function __construct( $args = array() ) {
 
 		// Set our default arguments, parse against the provided arguments, and store for use throughout the class.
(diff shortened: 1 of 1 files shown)


gravity-forms/gw-update-posts.php
```diff
@@ -97,9 +97,13 @@ public function update_post_by_entry( $entry, $form ) {
 		}
 
 		if ( $this->_args['post_date'] ) {
-			$new_date_time       = $this->get_post_date( $entry, $form );
-			$post->post_date     = $new_date_time;
-			$post->post_date_gmt = get_gmt_from_date( $new_date_time );
+			$new_date_time = $this->get_post_date( $entry, $form );
```
claygriffiths, line 100: Would it make more sense to return `null` in `get_post_date` if these aren't set?

gravity-forms/gw-update-posts.php
```diff
@@ -97,9 +97,13 @@ public function update_post_by_entry( $entry, $form ) {
...
+			if ( $this->_args['post_date']['date'] ) {
+				$post->post_date = $new_date_time;
+			}
+			if ( $this->_args['post_date']['time'] ) {
+				$post->post_date_gmt = get_gmt_from_date( $new_date_time );
+			}
```
claygriffiths, line 106: This `time` conditional doesn't make sense for GMT.

claygriffiths (commented):
I'd like to get @spivurno's thoughts on this as I'm not sure what the default behavior is supposed to be.

saifsultanc:
@claygriffiths @spivurno Sorry for missing this out. Arghh!

About the point you mentioned Dave, the `post_date` array is still set with the default args:

```
		$this->_args = wp_parse_args(
			$args,
			array(
                                 ....
				'post_date'       => array(
					'date' => false,
					'time' => false,
				),
```

So the code update proposed with this PR for the portion:
```
		if ( $this->_args['post_date'] ) {
			$new_date_time = $this->get_post_date( $entry, $form );
			$post->post_date = $new_date_time;
			$post->post_date_gmt = get_gmt_from_date( $new_date_time );
		}
```

In actuality, perhaps it may make more sense if we add the the IF more granularly:

`if ( $this->_args['post_date']['date'] && $this->_args['post_date']['time'] )`

But then my only concern was, what if the user is only setting the date and not the time (or vice-versa). That's why I proposed the separate checks. But yes, coming back to the main point I was targetting. Using `$this->_args['post_date']` as an initial check doesn't do what we need it to do.

And 'return null' with those checks on `get_post_date` would still try to fallback to the current date/time values. Tested it out as well.

Thoughts?

saifsultanc:
Updated as discussed on slack.

claygriffiths (approved):
LGTM! S&M since the commit messages are the same.

PR #747, opened 2024-01-02
`gw-advanced-merge-tags.php`: Fixed an compatibility issue with `order_summary` merge tag modifiers.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2463972101/59316?folderId=3808239

## Summary

The `order_summary:my_custom_template` doesn't work when the snippet `GW_Advanced_Merge_Tags` is activated.

The `GW_Advanced_Merge_Tags` snippet just needs to ensure its `gform_replace_merge_tags` logic doesn't happen too soon. The `gform_replace_merge_tags` happens at '10' (order summary and others), we make sure the advanced merge tags merge tag replacement happens later - say '12'.

**BEFORE:**
Screenshot 2024-01-02 at 8 43 58 AM

**AFTER:**
Screenshot 2024-01-02 at 8 43 42 AM

### DIFF
--- a/gravity-forms/gw-advanced-merge-tags.php
+++ b/gravity-forms/gw-advanced-merge-tags.php
@@ -78,7 +78,7 @@ class GW_Advanced_Merge_Tags {
 		add_action( 'gform_pre_render', array( $this, 'support_dynamic_population_merge_tags' ) );
 
 		add_action( 'gform_merge_tag_filter', array( $this, 'support_html_field_merge_tags' ), 10, 4 );
-		add_action( 'gform_replace_merge_tags', array( $this, 'replace_merge_tags' ), 10, 3 );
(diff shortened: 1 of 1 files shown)


claygriffiths (approved):
Ideally, some info on why `replace_merge_tags` is messing this up would be ideal, but since this snippet isn't wildly popular like All Fields Template, I think this is fine for now.

saifsultanc:
> Ideally, some info on why `replace_merge_tags` is messing this up would be ideal, but since this snippet isn't wildly popular like All Fields Template, I think this is fine for now.

Updated PR details. The `gform_replace_merge_tags` default logic at priority 10 is getting overridden which is deferred with the addition of priority 12 here.

PR #749, opened 2024-01-06
`gpnf-map-child-entries-to-acf-repeater-field.php`: Fixed an issue with nested form field data not displaying on ACF field.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2438225719/58343?folderId=7098280

## Summary

When the parent form is submitted, a post is created with the Advanced Post Creation add-on. The Nested Form entries should be linked to the targetted repeater field in the created post, but they don't work. This is down to the asynchronous behavior of Gravity Forms Advanced Post Creation Add-on v1.4.0 ([changelog](https://docs.gravityforms.com/advanced-post-creation-change-log/)). Disabling async does resolve the issue.

Also, the snippet is converted to a class-based snippet for easy management by the end user.

### DIFF
--- a/gp-nested-forms/gpnf-map-child-entries-to-acf-repeater-field.php
+++ b/gp-nested-forms/gpnf-map-child-entries-to-acf-repeater-field.php
@@ -5,7 +5,10 @@
  */
 
 class GPNF_Map_Child_Entries_To_ACF_Repeater {
-	function __construct( $args ) {
+
+	private $args;
+
+	public function __construct( $args ) {
 		$this->_args = wp_parse_args( $args, array(
 			'form_id'                 => false,
 			'nested_form_field_id'   => false,
--- a/gp-nested-forms/gpnf-map-child-entries-to-acf-repeater-field.php
+++ b/gp-nested-forms/gpnf-map-child-entries-to-acf-repeater-field.php
@@ -47,7 +47,7 @@ class GPNF_Map_Child_Entries_To_ACF_Repeater {
 	'form_id'                  => 7, // Set this to the parent form ID
 	'nested_form_field_id'     => 18, // Update to the ID of the Nested Form field.
 	'field_map'                => array(
-		'nom_comedien'         => 1,
+		'num_comedien'         => 1,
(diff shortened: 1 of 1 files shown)


spivurno:
@saifsultanc Rather than disabling async I'd explore the [gform_advancedpostcreation_post_after_creation](https://docs.gravityforms.com/gform_advancedpostcreation_post_after_creation/) filter instead. If we want to preserve compat for posts generated via Post fields, there's [https://docs.gravityforms.com/gform_after_create_post/](https://docs.gravityforms.com/gform_after_create_post/) for that scenario.

saifsultanc:
> @saifsultanc Rather than disabling async I'd explore the [gform_advancedpostcreation_post_after_creation](https://docs.gravityforms.com/gform_advancedpostcreation_post_after_creation/) filter instead. If we want to preserve compat for posts generated via Post fields, there's [gform_after_create_post](https://docs.gravityforms.com/gform_after_create_post/) for that scenario.

@spivurno Thanks for pointing out those hooks!

gp-nested-forms/gpnf-map-child-entries-to-acf-repeater-field.php
```diff
@@ -3,36 +3,53 @@
...
-		update_field( $acf_repeater_field_name, $repeat_value, $post_id );
 	}
+ }
 
-}
+ new GPNF_Map_Child_Entries_To_ACF_Repeater( array(
+	'form_id'                  => 7, // Set this to the parent form ID
+	'nested_form_field_id'     => 18, // Update to the ID of the Nested Form field.
+	'field_map'                => array(
+		'nom_comedien'         => 1,
```
claygriffiths, line 50:
```suggestion
		'nom_comedian'         => 1,
```

Not a blocker, but figured I'd point out this typo. It's in a few spots.

Should this also be `num` instead of `nom`? I just want to make sure this example makes sense.

claygriffiths (approved):
Looks good! Can you clean-up the usage before merging? There are some typos.

PR #751, opened 2024-01-09
`gw-all-fields-template.php`: Fixed a fatal error with the snippet and Nested Forms.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2460908266/59240?folderId=3808239

## Summary

The All Fields Template is causing a fatal error:
`Warning: Undefined variable $input_ids in /.../gw-all-fields-template/gw-all-fields-template.php on line 201`

The issue is down to the `$input_ids` only getting initialized in the preceding else block (and not the if). This leads to potential scenarios of tapping in on that value/variable even though it does not exist.

**BEFORE:**
Screenshot 2024-01-09 at 9 27 11 AM

**AFTER:**
Screenshot 2024-01-09 at 9 26 38 AM

### DIFF
--- a/gravity-forms/gw-all-fields-template.php
+++ b/gravity-forms/gw-all-fields-template.php
@@ -8,7 +8,7 @@
  * Plugin URI:   https://gravitywiz.com/gravity-forms-all-fields-template/
  * Description:  Modify the {all_fields} merge tag output via a template file.
  * Author:       Gravity Wiz
- * Version:      0.10
+ * Version:      0.11
  * Author URI:   http://gravitywiz.com
  *
  * Usage:
--- a/gravity-forms/gw-all-fields-template.php
+++ b/gravity-forms/gw-all-fields-template.php
@@ -175,6 +175,7 @@ class GW_All_Fields_Template {
 			 */
 			if ( $context == 'nested' ) {
 
+				$input_ids = array();
 				$nested_form_field_id = rgar( $modifiers, 'parent', false );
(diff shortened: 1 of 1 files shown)


gravity-forms/gw-all-fields-template.php
```diff
@@ -175,6 +175,7 @@ public function all_fields_extra_options( $value, $merge_tag, $modifiers, $field
 			 */
 			if ( $context == 'nested' ) {
 
+				$input_ids = array();
```
claygriffiths, line 178: Is the customer trying to get an input from the Nested Form? It seems like if this is the logic they're hitting we need to account for it in the foreach.

claygriffiths (commented):
See below

gravity-forms/gw-all-fields-template.php
claygriffiths: Disregard, I see that they're not using input IDs for now. It would be a nice improvement at some point.

claygriffiths (approved):
LGTM. Amend this last commit and increment the version here before merging: https://github.com/gravitywiz/snippet-library/blob/21ccea5cca0c94a6030d26566740dbc168a522f1/gravity-forms/gw-all-fields-template.php#L11C1-L11C1

saifsultanc:
@claygriffiths Ready to merge it. But the PHPCS issue pointing to a file that hasn't been touched on this PR.

PR #752, opened 2024-01-11
`gpadvs-enable-add-new-option.php`: Fixed an issue where adding a new option may result in failed validation during form submission.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2473904075/59646?folderId=3808239

## Summary

The [GPADVS Enable "Add New" Option](https://github.com/gravitywiz/snippet-library/blob/master/gp-advanced-select/gpadvs-enable-add-new-option.php) snippet is causing a validation error when GPPA is not enabled on a dropdown field.

We add logic for `gform_validation` to ensure we do not erroneously 'fail' for advanced select fields in these scenarios.

**BEFORE:**
https://www.loom.com/share/46e1cfeae5314b0fbc476b79bb4bdc5f

**AFTER:**
https://www.loom.com/share/39da118bd3d544e0a8d1a53d36ca1311


gp-advanced-select/gpadvs-enable-add-new-option.php
```diff
@@ -37,6 +37,7 @@ public function init() {
 
 		add_filter( 'gform_pre_render', array( $this, 'allow_created_choices_in_save_and_continue' ), 10, 3 );
 
+		add_filter( 'gform_validation', array( $this, 'validate' ) );
```
spivurno, line 40: Gravity Forms is strongly favoring the [gform_field_validation](https://docs.gravityforms.com/gform_field_validation/) filter these days. Let's roll with that one. It will simplify the `validate()` function as well.

saifsultanc: @spivurno Oh.. Okay. Will make the changes. Thanks for pointing that out :)

claygriffiths: This probably has to do with state validation. You can prevent state validation by setting `validateState` on the field to `false` during `gform_pre_render`.

We already do this some in GP Advanced Select when lazy-loading with GPPA: https://github.com/gravitywiz/gp-advanced-select/blob/34dc8d067dde0d5797979e004f97c9ba2cbc3fb7/class-gp-advanced-select.php#L388-L396

gp-advanced-select/gpadvs-enable-add-new-option.php
```diff
@@ -155,6 +160,15 @@ public function allow_created_choices_in_save_and_continue( $form, $ajax, $field
 		return $form;
 	}
 
+	public function disable_state_validation_for_advanced_select_field( $form ) {
+		foreach ( $form['fields'] as &$field ) {
```
claygriffiths, line 164: I'd add a check at the top of the method for `is_applicable_form`.

gp-advanced-select/gpadvs-enable-add-new-option.php
```diff
@@ -155,6 +160,15 @@ public function allow_created_choices_in_save_and_continue( $form, $ajax, $field
...
+			if ( $field->id == $this->_args['field_id'] ) {
```
claygriffiths, line 165: I think this logic is a bit too naïve as it won't work if you enable it for a form globally.

It might be time to move the following into an `is_applicable_field` method:

```
if ( ! gp_advanced_select()->is_advanced_select_field( $field ) ) {
    continue;
}

// Check if this instance is targeting all Advanced Select fields or a specific field.
if ( ! empty( $this->_args['field_id'] ) && $this->_args['field_id'] != $field->id ) {
    continue;
}
```

Source: https://github.com/gravitywiz/snippet-library/blob/4a6204341ee37244ad8fe6329c137afa564f8283/gp-advanced-select/gpadvs-enable-add-new-option.php#L128-L135

saifsultanc: @claygriffiths  Thanks for pointing out the mistake of tapping the form globally. I have updated as per your suggestions, also refactored to have a new method `is_applicable_field`

PR #755, opened 2024-01-16
`gw-first-error-focus.php`: Removed behavior to scroll to the top of the form if there is no error unless it's a multi-page form.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2479112475/59893?folderId=3808239
https://secure.helpscout.net/conversation/2482109891/60004?folderId=3808239

## Summary

The snippet takes the page to the `gform_wrapper` div on page load (i.e. when there are no rendered errors). The logic to scroll to `gform_wrapper` when there are no errors is the one that is causing the issues: `window.scrollTo(0, $('.gform_wrapper').offset().top);`

**Part - 1:**
https://www.loom.com/share/2a52775855d9413190f361d133a5e504
**Part - 2:**
https://www.loom.com/share/0aae06dc58374e3f86e6ff3cecbdbdd5

### DIFF
--- a/gravity-forms/gw-first-error-focus.php
+++ b/gravity-forms/gw-first-error-focus.php
@@ -35,7 +35,7 @@ function gw_first_error_focus_script() {
 
 					if (!hasError) {
 						requestAnimationFrame(function() {
-							window.scrollTo(0, $('.gform_wrapper').offset().top);
+							window.scrollTo(0, 0);
 						});
 					}
 				});
--- a/gravity-forms/gw-first-error-focus.php
+++ b/gravity-forms/gw-first-error-focus.php
@@ -31,11 +31,16 @@ function gw_first_error_focus_script() {
 					// We need to reset our flag so that we can still do our focus action when the form conditional logic
 					// has been re-evaluated.
 					window['gwfef'] = false;
+					var y = 0;
+					// For a multiple page form after page 1, we need to render on the gform wrapper
+					if ($('.gf_step_current_page') && $('.gf_step_current_page').text() > 1) {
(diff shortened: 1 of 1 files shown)


gravity-forms/gw-first-error-focus.php
```diff
@@ -31,11 +31,16 @@ function gw_first_error_focus_script() {
 					// We need to reset our flag so that we can still do our focus action when the form conditional logic
 					// has been re-evaluated.
 					window['gwfef'] = false;
+					var y = 0;
+					// For a multiple page form after page 1, we need to render on the gform wrapper
+					if ($('.gf_step_current_page') && $('.gf_step_current_page').text() > 1) {
+						y = $('.gform_wrapper').offset().top;
+					}
 					var hasError = gwFirstErrorFocus().length;
```
claygriffiths, line 39: I'd create a new variable below `hasError` named `onSubsequentPage` or something like that...

`var onSubsequentPage = $('.gf_step_current_page') && $('.gf_step_current_page').text() > 1;`

Then, change `if (!hasError) {` to `if (!hasError && onSubsequentPage) {`

All other modifications aren't needed, I don't think.

The one caveat with this is if they navigate from the second page to the first, it won't work, but I can't think of a quick way to handle that.

claygriffiths:
Related PR: https://github.com/gravitywiz/snippet-library/pull/477

I added notes here: https://github.com/gravitywiz/snippet-library/pull/477#issuecomment-1898814854

claygriffiths (approved):
LGTM.

Squash & Merge with the updated PR title

PR #756, opened 2024-01-17
`gw-export-multi-input-fields-single-colum.php`: Fixed a compatibility issue with GP Google Sheets.

saifsultanc:
## Context
⛑️ Ticket(s): https://secure.helpscout.net/conversation/2471019694/59531?folderId=3808239

## Summary

Checkbox fields are not sent to Google Sheets when Gravity Forms - Export Multi-input Fields in Single Column plugin is active.

The proposed solution in this PR is to skip the `gform_export_fields` hook's processing when the trigger is `GP Google Sheets`, by checking on the `$_POST` object.

### DIFF
--- a/gravity-forms/gw-export-multi-input-fields-single-colum.php
+++ b/gravity-forms/gw-export-multi-input-fields-single-colum.php
@@ -10,15 +10,10 @@
  * Plugin URI: http://gravitywiz.com/how-do-i-export-multi-input-fields-in-a-single-column-with-gravity-forms/
  * Description: Export multi-input Gravity Forms fields as a single column.
  * Author: David Smith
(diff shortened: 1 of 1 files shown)


claygriffiths:
For whatever reason, I was unable to reproduce this when trying to recreate the issue in my development site. I was able to recreate it using the backup, however.

I noticed that there is a better hook to use (`gform_form_export_page`), so I switched to that to avoid the need for the conditional.

GPGS is using `gform_form_export_page` when inserting rows, so I've created the following PR which removes that call: https://github.com/gravitywiz/gp-google-sheets/pull/94

Between these two PRs, the issue is resolved for me in the backup used to recreate the issue.

PR #757, opened 2024-01-17
`gppa-acf-repeater-mapper.php`: Fixed an issue with regular expression.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2470092230/59496?folderId=3808239

## Summary

A limitation with the [ACF Repeater Mapper snippet](https://github.com/gravitywiz/snippet-library/blob/master/gp-populate-anything/gppa-acf-repeater-mapper.php) if the repeater field has a number attached to the name.

So, if we processed for `$key` as `meta_test_repeater_1_0_name`. This will give us:

`$repeater` as `test_repeater`
`$index` as `1`
`$subfield` as `0_name`

This is because of the additional 1 attached to the name. The `$repeater` is actually `test_repeater_1`, `$index` is `0`, and `$subfield` is `name`.


claygriffiths (approved):
This regex LGTM!

CleanShot 2024-01-17 at 10 01 22

PR #759, opened 2024-01-18
`gw-field-to-field-conditional-logic.php`: Fixed an issue with number formatting.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2477359501/59822?folderId=3808239

## Summary

The snippet above doesn't work if the result of a calculation field used for the conditional logic check is in the thousands.

The issue is with comparing numeric values that are comma formatted.

**BEFORE:** ('Show This' did not hide even if the values 1000 and 1,000 are numerically equal.)
Screenshot 2024-01-18 at 12 32 12 PM

**AFTER:**
Screenshot 2024-01-18 at 12 31 49 PM

### DIFF
--- a/gravity-forms/gw-field-to-field-conditional-logic.php
+++ b/gravity-forms/gw-field-to-field-conditional-logic.php
@@ -138,9 +138,22 @@ class GF_Field_To_Field_Conditional_Logic {
 							}
 
 							rule.value = GFMergeTag.getMergeTagValue( formId, mergeTags[0][1], mergeTags[0][3] );
+							rule.value = convertIfNumberWithCommas( rule.value );
 
 							return rule;
 						} );
+
+						function convertIfNumberWithCommas( str ) {
+							// Regex to check if the string is a number with commas.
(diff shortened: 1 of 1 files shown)


gravity-forms/gw-field-to-field-conditional-logic.php
```diff
@@ -138,10 +138,23 @@ public function output_script() {
 							}
 
 							rule.value = GFMergeTag.getMergeTagValue( formId, mergeTags[0][1], mergeTags[0][3] );
+							rule.value = convertIfNumberWithCommas( rule.value );
```
claygriffiths, line 141: Here's how GF does this, I'd lean that direction:

```js
var fieldNumberFormat = gf_get_field_number_format( rule.fieldId, formId, 'value' );

if( fieldNumberFormat ) {
    fieldValue = gf_format_number( fieldValue, fieldNumberFormat );
}
```

claygriffiths (approved):
LGTM!

PR #760, opened 2024-01-19
`gpnf-auto-attach-child-files.php`: Added support for Image Hopper Files.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2481484636/59962?folderId=3808239

## Summary

This snippet is supposed to auto-attach child files to the notification of the parent form. It works for the GF File upload fields, it doesn't work when an Image Hopper file upload field. This update adds support for that. We need to be specifying what upload files we need from the child entries:

`$upload_fields = GFCommon::get_fields_by_type( $child_form, array( 'fileupload', 'image_hopper', 'image_hopper_post' ) );`

### DIFF
--- a/gp-nested-forms/gpnf-auto-attach-child-files.php
+++ b/gp-nested-forms/gpnf-auto-attach-child-files.php
@@ -71,7 +71,7 @@ class GPNF_Auto_Attach_Child_Files {
 			return $notification;
 		}
 
-		$upload_fields = GFCommon::get_fields_by_type( $form, array( 'fileupload' ) );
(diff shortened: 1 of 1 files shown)


saifsultanc:
@claygriffiths It's only Image Hopper right now, but I wonder if it would a bit more practical sense if we added extra "upload" field types when we create the object. So something like:

'additional_upload_field_types' => array( 'image_hopper', 'image_hopper_post' ),

claygriffiths:
@saifsultanc Love the idea! My only worry is that the data structure of the other upload fields may not match what we're expecting.

claygriffiths (approved):

PR #764, opened 2024-01-25
`gppa-acf-repeater-mapper.php`: Fixed an issue with regular expression.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2454263207/58976/

## Summary

Follow up to this previous update https://github.com/gravitywiz/snippet-library/pull/757 for the issue reported [here](https://secure.helpscout.net/conversation/2470092230/59496?folderId=3808239). We are rolling back to the previous state, adding a comment that we cannot use numbers in repeater names as a known limitation.

### DIFF
--- a/gp-populate-anything/gppa-acf-repeater-mapper.php
+++ b/gp-populate-anything/gppa-acf-repeater-mapper.php
@@ -36,7 +36,8 @@ add_filter( 'gppa_input_choices', function( $choices, $field, $objects ) {
 
 	foreach ( $field->{'gppa-choices-templates'} as $template => $key ) {
 		// Look for ACF repeater meta: repeater_name_0_subfield_name
-		if ( preg_match( '/^meta_(.*?)_([^_]+)_([^_]+)$/', $key, $matches ) ) {
+		// Known Limitation: We cannot have a number on the repeater field name.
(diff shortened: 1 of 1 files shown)


gp-populate-anything/gppa-acf-repeater-mapper.php
```diff
@@ -36,7 +36,8 @@
 
 	foreach ( $field->{'gppa-choices-templates'} as $template => $key ) {
 		// Look for ACF repeater meta: repeater_name_0_subfield_name
-		if ( preg_match( '/^meta_(.*?)_([^_]+)_([^_]+)$/', $key, $matches ) ) {
+		// Known Limitation: We cannot have a number on the repeater field name.
```
claygriffiths, line 39: Thanks for the comment—can you expand on this and add an example of what will work and what won't work?

claygriffiths (approved):

gp-populate-anything/gppa-acf-repeater-mapper.php
spivurno: I'd also add this known limitation in the instructions so users can catch this on their own. Also, I mentioned that we could suggest using "a, b, c" if they need to use an enumerator in their field names.

PR #766, opened 2024-01-30
`gw-prevent-duplicate-selections.php`: Fixed an issue with duplicate selections logic did not work with some dynamically populated data.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2493953931/60560?folderId=3808239

## Summary

The snippet doesn't work when the dropdown field choices are not set when the form renders. If choices are populated later based on another field selection, it doesn't work. We need to ensure to re-render the logic with `gppa_updated_batch_fields` for updated fields.

A quick summary screencast of testing this update:
https://www.loom.com/share/a668853d694b4575b0132a077eeeb00d

### DIFF
--- a/gravity-forms/gw-prevent-duplicate-selections.php
+++ b/gravity-forms/gw-prevent-duplicate-selections.php
@@ -47,7 +47,7 @@ class GW_Prevent_Duplicate_Selections {
 
 		<script type="text/javascript" defer>
 
-			jQuery( function( $ ) {
+			function process_duplicate_selections($) {
 				window.gform.addFilter( 'gplc_excluded_input_selectors', function( selectors ) {
 					selectors.push( '.gw-disable-duplicates-disabled' );
 					return selectors;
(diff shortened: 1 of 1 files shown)


claygriffiths:
@saifsultanc I was waiting on a customer to reply before merging https://github.com/gravitywiz/snippet-library/pull/762 but I went ahead and merged it since it will conflict like crazy with this PR. 

With the latest version of the snippet, if you put the GPPA listener inside the `window.<?php echo __CLASS__; ?> = function() {` and call `gwDisableDuplicates` accordingly on the freshed fields, I think that'll do.

gravity-forms/gw-prevent-duplicate-selections.php
```diff
@@ -47,7 +47,7 @@ public function output_script() {
 
 		<script type="text/javascript" defer>
 
-			jQuery( function( $ ) {
+			function process_duplicate_selections($) {
```
spivurno, line 50:
```suggestion
			function gwProcessDuplicateSelections($) {
```

We typically use camel case for JS functions. I also prefixed this with `gw` since I believe this will be in the global scope?

claygriffiths (commented):
This looks solid. It looks like you need to rebase this with https://github.com/gravitywiz/snippet-library/blob/master/gravity-forms/gw-prevent-duplicate-selections.php as there is currently a merge conflict.

You might have to start a new branch based off of master and bring in your changes manually.

More specifically, it looks like you need to bring in the `gform_post_conditional_logic`  behavior, the `gppa_updated_batch_fields` addition, and whatever else you see fit.

PR #769, opened 2024-02-13
`gw-cache-buster.php`: Improved support for `{embed_url}` and updating the `source_url` property on the entry during submission.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2505977055/61172?folderId=7098280

## Summary

When using the Cache Buster, the `{embed_url}` merge tag returns a gfcb URL instead of the standard URL.

**Steps to Reproduce**
1. Add `{embed_url}` to your form notification or confirmation.
2. Embed the form on your site using Cache Buster.
3. Submit an entry.
4. Check notification, URL is different than normal. Example: `https://yoursite.local/wp-admin/admin-ajax.php?action=gfcb_get_form&form_id=1302`

To fix the issue we add the `gform_pre_replace_merge_tags` to check for `{embed_url}` merge tag and ensure the original link continues to be sent forth, and not the gfcb cache builder link.

### DIFF
--- a/gravity-forms/gw-cache-buster.php
+++ b/gravity-forms/gw-cache-buster.php
@@ -141,11 +141,12 @@ class GW_Cache_Buster {
 		// Include original query parameters (with some exclusions) in the AJAX call to preserve dynamic population via query string.
 		$exclude_params = array( 'action', 'form_id', 'atts' );
 		$ajax_url       = remove_query_arg( $exclude_params, add_query_arg( $_GET, admin_url( 'admin-ajax.php' ) ) );
-		$ajax_url       = add_query_arg(
+
+		// Still needed for the AJAX submission.
+		$ajax_url = add_query_arg(
 			array(
-				'action'              => 'gfcb_get_form',
-				'form_id'             => $form_id,
-				'form_request_origin' => rgpost( 'gwcb_form_request_origin' ),
+				'action'  => 'gfcb_get_form',
+				'form_id' => $form_id,
 			),
 			$ajax_url
 		);
@@ -162,7 +163,8 @@ class GW_Cache_Buster {
 				$.post( '<?php echo $ajax_url; ?>', {
 					action: 'gfcb_get_form',
 					form_id: '<?php echo $form_id; ?>',
-					atts: '<?php echo json_encode( $attributes ); ?>',
+					atts: '<?php echo esc_js( json_encode( $attributes ) ); ?>',
+					form_request_origin: '<?php echo esc_js( GFCommon::openssl_encrypt( GFFormsModel::get_current_page_url() ) ); ?>',
 					lang: '<?php echo $lang; ?>'
 				}, function( response ) {
 					$( '#gf-cache-buster-form-container-<?php echo $form_id; ?>' ).html( response ).fadeIn();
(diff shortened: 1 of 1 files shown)


gravity-forms/gw-cache-buster.php
```diff
@@ -271,6 +272,25 @@ public function filter_resume_link( $resume_url, $form, $resume_token, $email )
...
+		if ( strpos( $text, '{embed_url}' ) !== false ) {
+	
+			$url        = $_SERVER['REQUEST_URI'];
+			$parsed_url = parse_url($url);
+	
+			parse_str( $parsed_url['query'], $query_params );
+	
+			$form_request_origin = rgar( $query_params, 'form_request_origin' );
+			$decoded_value       = urldecode($form_request_origin);
+			$original_url        = $_SERVER['HTTP_ORIGIN'] . $decoded_value;
```
claygriffiths, line 286: Can this all be simplified to the following?

```suggestion
			$original_url = rgget( 'form_request_origin' );
```

claygriffiths (changes requested):
I think this approach is halfway there as it would work if you put `{embed_url}` in a field's default value, but it feels a bit flaky for notifications/confirmations.

Typically, Gravity Forms uses `$entry['source_url']` after a submission. We should hook into the after submission hook and update `source_url` on the entry accordingly.

The one thing I am a bit unsure on is this `replace_embed_url` is currently global. I'm wondering how we make sure non Cache Buster forms are not affected.

saifsultanc:
works good. thanks!

gravity-forms/gw-cache-buster.php
claygriffiths: @saifsultanc Sorry, this was a bad suggestion. I didn't realize that `form_request_origin` was `$_SERVER['REQUEST_URI']`.

```php
$form_request_origin = rgar( $query_params, 'form_request_origin' );
$decoded_value         = urldecode( $form_request_origin );
$origin_url                   = $_SERVER['HTTP_ORIGIN'] . $decoded_value;
```

... does make sense here, otherwise it'll just be the path.

gravity-forms/gw-cache-buster.php
```diff
@@ -271,6 +272,18 @@ public function filter_resume_link( $resume_url, $form, $resume_token, $email )
 		return add_query_arg( array( 'gf_token' => $resume_token ), $referer );
 	}
 
+	public function replace_embed_url( $text, $form, $entry, $url_encode ) {
+		// Check if the text contains the {embed_url} merge tag
+		if ( strpos( $text, '{embed_url}' ) !== false ) {
+	
+			$original_url = rgget( 'form_request_origin' );
```
claygriffiths, line 279: @saifsultanc It seems like this will still only impact default values. 

Does it work after submission and show properly in the entry details? I think we'll need to add a hidden field to track this. The hidden field will contain what is being populated in `form_request_origin`.

Here's an example of how we do that: 

https://github.com/gravitywiz/gp-entry-blocks/blob/master/includes/blocks/class-edit-form.php#L187-L205

It's hooked here: https://github.com/gravitywiz/gp-entry-blocks/blob/master/includes/blocks/class-edit-form.php#L87

Lastly, I do think security is a concern here. Because cache buster is enabled using a shortcode, we don't really know if it's being used for a form or not. 

Because of this, I recommend encrypting the hidden field value using `GFCommon::openssl_encrypt()` and then decrypted using `GFCommon::openssl_decrypt()`. If decryption fails, we just skip modifying `source_url`. It's probably worth adding this same treatment to `form_request_origin` when we're building `$ajax_url`.

claygriffiths (commented):
One more pass. Let me know if you need any more pointers.

gravity-forms/gw-cache-buster.php
saifsultanc: **CONFIRMATION PAGE:** ✅
Screenshot 2024-02-28 at 10 45 57 AM

**NOTIFICATION PAGE:** ✅
Screenshot 2024-02-28 at 10 46 11 AM

**ENTRY VIEW:** ✅
Screenshot 2024-02-28 at 10 46 57 AM

saifsultanc: Loom: https://www.loom.com/share/34272213f3944d2ea67d8d321c858e53

PR #778, opened 2024-03-08
Fix the inability to remove decimal quantity from the cart

lanresmith:
## Context

## Summary
The current snippet can only remove integer quantities from the cart. It is however possible to have decimal quantity especially with the use of the `gravity-forms/gw-quantity-as-decimal.php` snippet.


spivurno:
Closing due to no attached ticket and unclear usage.

PR #780, opened 2024-03-12
`gw-cache-buster.php:` Fixed an issue where some shortcode attributes are not passed to `gravity_form` function.

barthc:
⛑️ Ticket(s): https://secure.helpscout.net/conversation/2535157762/63075/

## Summary
Some shortcode  attributes are not passed to the `gravity_form` function, and that is due to the `esc_js` causing the `json_decode` to return NULL.

### DIFF
--- a/gravity-forms/gw-cache-buster.php
+++ b/gravity-forms/gw-cache-buster.php
@@ -163,7 +163,7 @@ class GW_Cache_Buster {
 				$.post( '<?php echo $ajax_url; ?>', {
 					action: 'gfcb_get_form',
 					form_id: '<?php echo $form_id; ?>',
-					atts: '<?php echo wp_json_encode( $attributes ); ?>',
+					atts: <?php echo wp_json_encode( $attributes ); ?>,
 					form_request_origin: '<?php echo esc_js( GFCommon::openssl_encrypt( GFFormsModel::get_current_page_url() ) ); ?>',
 					lang: '<?php echo $lang; ?>'
 				}, function( response ) {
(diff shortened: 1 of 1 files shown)


gravity-forms/gw-cache-buster.php
```diff
@@ -163,7 +163,7 @@ public function shortcode( $markup, $attributes, $content ) {
 				$.post( '<?php echo $ajax_url; ?>', {
 					action: 'gfcb_get_form',
 					form_id: '<?php echo $form_id; ?>',
-					atts: '<?php echo esc_js( json_encode( $attributes ) ); ?>',
+					atts: '<?php echo wp_json_encode( $attributes ); ?>',
```
spivurno, line 166: @saifsultanc I see you recently introduced the `esc_js()` call here: https://github.com/gravitywiz/snippet-library/commit/a0dbcae636ddad15d5661d4d76344bc47c72a652#diff-f0e934343c5ea8212d135a48c8f339f8f45d298ae171f51f563b1d245cd2d424R166

Why was this needed?

saifsultanc: @spivurno Not sure, I would want to know myself! This bit was actually committed by Clay: https://github.com/gravitywiz/snippet-library/pull/769/commits/6edd92305a8549a3a7438224a2c2e0d5eacd0e56

spivurno: @claygriffiths `esc_js()` is causing issues here. Do you remember what issues it resolved?

claygriffiths: I don't recall there being a specific issue—just a situation where it felt wrong to be putting JSON in a string. It's one apostrophe away from breaking.

I originally got rid of the single quotes on `atts`, but the PHP side was JSON decoding it and I stopped there and threw on the `esc_js`.

claygriffiths: [snippet-library-9-43-52 AM.patch](https://github.com/gravitywiz/snippet-library/files/14588521/snippet-library-9-43-52.AM.patch)

This does work, I should've just made the change. 😅

Tested shortcode:

```
[gravityform id="4" cachebuster="true" description="false" title="false"]
```

PR #781, opened 2024-03-13
`gpadvs-preserve-choice-order.php`: Added support to preserve choice order with Multi Select fields using GP Advanced Select.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2522300498/62231?folderId=3808239

## Summary

With a Multi-Select field, using GP Advanced Select, selected choices are stored in the entry in the order that they are selected. This snippet helps override that behaviour to preserve the order choice.

Screencast:
https://www.loom.com/share/d2ad3fea02234f02871ac9b1efe20b53

### DIFF
--- a/gp-advanced-select/gpadvs-preserve-choice-order.php
+++ b/gp-advanced-select/gpadvs-preserve-choice-order.php
@@ -1,0 +1,35 @@
+<?php
+/**
+ * Gravity Perks // Advanced Select // Preserve Choice Order
+ * https://gravitywiz.com/documentation/gravity-forms-advanced-select/
+ *
+ * Preserve Choice Order with Multi Select Fields using GP Advanced Select.
+ * 
+ * Instruction Video: https://www.loom.com/share/d2ad3fea02234f02871ac9b1efe20b53
+ *
+ * Instructions:
+ *
+ * 1. Install this snippet by following the instructions here:
+ *    https://gravitywiz.com/documentation/how-do-i-install-a-snippet/
+ */
+add_filter( 'gform_save_field_value', 'preserve_choice_order_with_advanced_select', 10, 4 );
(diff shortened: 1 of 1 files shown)


gp-advanced-select/gpadvs-preserve-choice-order.php
```diff
@@ -0,0 +1,35 @@
+<?php
+/**
+ * Gravity Perks // Advanced Select // Preserve Choice Order
+ * https://gravitywiz.com/documentation/gravity-forms-advanced-select/
+ *
+ * Preserve Choice Order with Multi Select Fields using GP Advanced Select.
```
spivurno, line 6:
```suggestion
 * Preserve choice order with Multi Select fields using GP Advanced Select.
```

gp-advanced-select/gpadvs-preserve-choice-order.php
```diff
@@ -0,0 +1,35 @@
...
+ *
+ * Instruction Video: https://www.loom.com/share/d2ad3fea02234f02871ac9b1efe20b53
+ *
+ * Instructions:
+ *
+ * 1. Install this snippet by following the instructions here:
+ *    https://gravitywiz.com/documentation/how-do-i-install-a-snippet/
+ */
+add_filter( 'gform_save_field_value', 'preserve_choice_order_with_advanced_select', 10, 4 );
+function preserve_choice_order_with_advanced_select( $value, $lead, $field, $form ) {
```
spivurno, line 16: As a general rule, we use anonymous functions for the Snippet Library. My thinking is that the less the user has read in the code, the easier it is for them to parse and therefore less intimidating.

spivurno (approved):
Approved with requested changes. No need for re-review.

PR #782, opened 2024-03-14
`gw-advanced-merge-tags.php`: Added support for Live Merge Tags in HTML field to work on GP Post Content Merge Tags.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2535995803/63104?folderId=7098280

## Summary

The live merge tag in an HTML field isn't processed when the HTML field content is displayed on the confirmation page using the GPPCMT

BEFORE (Samuel):
https://www.loom.com/share/9f41bc1e25e9448fa083597ba363e203?sid=14eb078b-48cd-4425-a4d9-fa706a0826ba

AFTER:
https://www.loom.com/share/3480a1c24fc34897b0b211663d5f4137

### DIFF
--- a/gravity-forms/gw-advanced-merge-tags.php
+++ b/gravity-forms/gw-advanced-merge-tags.php
@@ -432,6 +432,17 @@ class GW_Advanced_Merge_Tags {
 	public function support_html_field_merge_tags( $value, $tag, $modifiers, $field ) {
 		if ( $field->type == 'html' && ( $tag != 'all_fields' || in_array( 'allowHtmlFields', explode( ',', $modifiers ) ) ) ) {
 			$value = $field->content;
+
+			// Support for LMTs with GP Post Content Merge Tags
+			if ( class_exists( 'GP_Populate_Anything_Live_Merge_Tags' ) && is_callable( 'gp_post_content_merge_tags' ) && rgget( 'eid' ) ) {
+				$gppa_lmt = GP_Populate_Anything_Live_Merge_Tags::get_instance();
+				if ( $gppa_lmt->has_live_merge_tag( $value ) ) {
+					$form  = GFAPI::get_form( $field->formId );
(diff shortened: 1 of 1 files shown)


spivurno (changes requested):
Mixed feelings on this one. The solution is good, I'm just torn up about placement.

It feels strange that Advanced Merge Tags should be responsible for integrating the functionality of two separate perks.

I wonder if this would be better placed in Post Content Merge Tags. Something like this.

```php
add_action( 'gform_merge_tag_filter', function( $value, $tag, $modifiers, $field ) {

	/**
	 * Merge tags for HTML fields (e.g. {My HTML Field:1}) are not supported by default. Our Advanced Merge Tags snippet
	 * adds support for these merge tags but there are other snippets out in the wild as well. If an HTML field merge tag
	 * has a value here, we'll assume it will be displayed and parse it for Live Merge Tags.
	 */
	if ( $field->type !== 'html' || empty( $value ) || ! class_exists( 'GP_Populate_Anything_Live_Merge_Tags' ) ) {
		return $value;
	}

	$gppa_lmt = GP_Populate_Anything_Live_Merge_Tags::get_instance();
	if ( ! $gppa_lmt->has_live_merge_tag( $value ) ) {
		return $value;
	}

	$entry = gp_post_content_merge_tags()->get_entry();
	if ( ! $entry ) {
		return $value;
	}

	$form = GFAPI::get_form( $field->formId );
	$gppa_lmt->populate_lmt_whitelist( $form );

	return $gppa_lmt->replace_live_merge_tags_static( $value, $form, $entry );
}, 10, 4 );```

Obviously, we wouldn't want to use an anonymous function in that context. 😄

saifsultanc:
Closing in favour of https://github.com/gravitywiz/gp-post-content-merge-tags/pull/21

PR #784, opened 2024-03-15
`gpadvs-preserve-choice-order.php`: Fixed PHP warnings.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2522300498/62231?folderId=7098280

## Summary

Fix PHP warnings with the snippet.

```
Warning: array_flip(): Can only flip STRING and INTEGER values!
Warning: usort() expects parameter 1 to be array, null given
```

### DIFF
--- a/gp-advanced-select/gpadvs-preserve-choice-order.php
+++ b/gp-advanced-select/gpadvs-preserve-choice-order.php
@@ -23,15 +23,11 @@ add_filter( 'gform_save_field_value', function ( $value, $lead, $field, $form ) 
 
 	// Create a map of choice values to their respective index, and sort with that.
 	if ( is_array( $field->choices ) && ! empty( $field->choices ) && $value_array ) {
-		$valid_values = array_filter( array_column( $field->choices, 'value' ), function ( $value ) {
-			return is_string( $value ) || is_int( $value );
-		});
-		$value_indices = array_flip( $valid_values );
-		if ( is_array( $value_indices ) ) {
-			usort( $value_array, function ( $x, $y ) use ( $value_indices ) {
-				return $value_indices[ $x ] - $value_indices[ $y ];
-			});
-		}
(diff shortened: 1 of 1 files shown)


gp-advanced-select/gpadvs-preserve-choice-order.php
```diff
@@ -22,10 +22,17 @@
...
+		$valid_values = array_filter( array_column( $field->choices, 'value' ), function ( $value ) {
+			return is_string( $value ) || is_int( $value );
+		});
+		$value_indices = array_flip( $valid_values );
+		if ( is_array( $value_indices ) ) {
+			usort( $value_array, function ( $x, $y ) use ( $value_indices ) {
+				return $value_indices[ $x ] - $value_indices[ $y ];
+			});
+		}
+	}
```
spivurno, line 35:
```suggestion
	if ( is_array( $field->choices ) && ! empty( $field->choices ) && $value_array ) {
		// Avoid PHP notices for non-string, non-integer values with `array_flip()` by casting everything to a string and ensuring the result is an array.
		$value_indices = (array) array_flip( array_map( 'strval', array_column( $field->choices, 'value' ) ) );
		usort( $value_array, function ( $x, $y ) use ( $value_indices ) {
			return $value_indices[ $x ] - $value_indices[ $y ];
		} );
	}
```

Thoughts on simplifying the code to something like this? Also added a comment about why we're gettin' fancy.

spivurno (approved):
Take a look at the suggested change. If you like it, send it to the customer for confirmation that it still prevents the notices they were experiencing. If so, consider this pre-approved. ✅

gp-advanced-select/gpadvs-preserve-choice-order.php
saifsultanc: Confirmed by. customer here: https://secure.helpscout.net/conversation/2522300498/62231?folderId=7098280#thread-7632446291

PR #785, opened 2024-03-16
Fix validation error for the decimal quantity field

lanresmith:
Fix validation error for the decimal quantity field when the value starts with the decimal sign (dot/comma).

As can be seen in the screenshot below, form validation fails for a value like `.2` - which is a valid value - even when the total is correctly calculated. This is quite inconsistent.

Code formatting/refactoring was also done for some parts of the existing code in order to provide clarity and reduce repetitions e.g. the `GFCommon::is_numeric` method was used 4 times in the old code but just twice in the new code.

Screenshot 2024-03-15 at 15 20 57

## Context


spivurno:
Can't find the ticket and haven't had any other reports of this.

PR #788, opened 2024-03-22
`gw-limit-multi-selects.php`: Fixed issue where setting the `min` to `1` did not work.

veryspry:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2499159955/60820/ 

## Summary

If no options are selected for a configured multi-select, the `$value` arg in `validate()` will be an empty string and the function will return early without setting a validation error.

The fix is to explicity set `$count` to `1` if an empty string is detected so that the remainder of the logic will work as expected.


claygriffiths (approved):
LGTM!

PR #789, opened 2024-03-22
`gw-update-posts.php`: Fixed issue where an empty post ID would result in the current post getting updated.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2545475210/63599/

## Summary

When using the [Update Posts](https://gravitywiz.com/how-to-update-posts-with-gravity-forms/) snippet, if the "Post ID" field is blank when the form is submitted, then the snippet will update the post that the form is embedded on, instead of just bailing.

**BEFORE**:
https://www.loom.com/share/185dd99a886c4194b0e59f677e2e8cd0

**AFTER**:
https://www.loom.com/share/2f655d3746a14659ba24ed6388c1bcb4

### DIFF
--- a/gravity-forms/gw-update-posts.php
+++ b/gravity-forms/gw-update-posts.php
@@ -68,8 +68,13 @@ class GW_Update_Posts {
 
 	public function update_post_by_entry( $entry, $form ) {
 
+		$post_id = rgar( $entry, $this->_args['post_id'] );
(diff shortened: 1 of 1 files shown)


claygriffiths (approved):
LGTM! Can you update the `@version` at the top of the file then S&M?

PR #795, opened 2024-03-29
`gpeb-disable-conditional-logic.php`: Added snippet to disable Field Conditional Logic when editing using Entry Blocks.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2552427358/63973/

## Summary

Snippet that removes conditional logic from all fields when editing an entry via Entry Blocks.


claygriffiths (approved):
Solid!

PR #796, opened 2024-03-31
`gw-all-checkboxes-checked.js`: Added snippet to trigger conditional logic evaluation when all checkboxes are checked.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2549746277/63832?folderId=7098280#

## Summary

This snippet populates a Hidden field with the value of 1 if every checkbox in a given field is checked and a 0 if every checkbox in that field is not checked.

Screencast:
https://www.loom.com/share/5c55fe7fa415428aa48b249d6ead071f


spivurno (approved):
Thank you!

PR #797, opened 2024-03-31
`gpadvs-auto-select-on-match.js`: Added new snippet for GP Advanced Select.

lanresmith:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2546318031/63648?folderId=8181748

## Summary

The newly-added snippet automatically selects the (single) matching option as the user types to filter the options in an enhanced select field.

### DIFF
--- a/gp-advanced-select/gpadvs-auto-select-on-match.js
+++ b/gp-advanced-select/gpadvs-auto-select-on-match.js
@@ -1,0 +1,31 @@
+/**
+ * Gravity Perks // Advanced Select // Auto Select on Match
+ * https://gravitywiz.com/documentation/gravity-forms-advanced-select/
+ *
+ * Automatically select the matching option based on the search value.
+ *
+ * Instructions:
+ *
+ * 1. Install this snippet with our free Custom JavaScript plugin.
+ *    https://gravitywiz.com/gravity-forms-code-chest/
+ */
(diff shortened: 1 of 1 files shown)


gp-advanced-select/gpadvs-auto-select-on-match.js
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * 1. Install this snippet with our free Custom JavaScript plugin.
+ *    https://gravitywiz.com/gravity-forms-code-chest/
+ */
+gform.addFilter( 'gpadvs_settings', function( settings, gpadvs, selectNamespace ) {
+  if ( gpadvs.formId == GFFORMID ) {
+    settings.onType = function( str ) {
+      const tomSelect = window[ selectNamespace ];
+      const matchedKey = Object.keys( tomSelect.options ).filter( function( key ) {
+        return -1 !== tomSelect.options[ key ].text.toLowerCase().indexOf( str.toLowerCase() );
```
claygriffiths, line 17: I think this is a decent enough stop-gap. I am mildly curious if we can use the list of results that are fuzzy-matched as the fuzzy-match might not always match this search method. Don't worry about digging into it right now, I'm putting this note here for future reference.

claygriffiths (approved):
Approval pending https://github.com/gravitywiz/gp-advanced-select/pull/29#pullrequestreview-1972150747

PR #801, opened 2024-04-04
`gw-create-coupon.php`: Added the ability to create coupons only when certain fields are not empty.

lanresmith:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2555639705/64138?folderId=8181748

### DIFF
--- a/gravity-forms/gw-create-coupon.php
+++ b/gravity-forms/gw-create-coupon.php
@@ -4,7 +4,7 @@
  *
  * Create coupons via Gravity Forms submissions. Map the coupon code to a field on the GF form and voila!
  *
- * @version 1.2.1
+ * @version 1.2.2
  * @author  David Smith <david@gravitywiz.com>
  * @license GPL-2.0+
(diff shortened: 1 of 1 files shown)


gravity-forms/gw-create-coupon.php
```diff
@@ -238,6 +239,22 @@ function is_applicable_form( $form ) {
...
 
+	public function should_abort( $entry ) {
+		if ( empty( $this->_args['required_fields'] ) ) {
+			return false;
+		}
+
+		foreach ( $this->_args['required_fields'] as $field_id ) {
+			$value = rgar( $entry, (string) $field_id );
+
+			if ( empty( $value ) ) {
```
claygriffiths, line 250: `rgblank` may be a better option here, otherwise entering `0` will not satisfy the field requirement.

lanresmith: @claygriffiths Fixed. Thanks!

PR #805, opened 2024-04-12
`gspc-add-order-meta-to-gf-entry.php`: Added a new snippet that adds order metadata to GF entries on export.

lanresmith:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2556313259/64159?folderId=8181748

### DIFF
--- a/gs-product-configurator/gspc-add-order-meta-to-gf-entry.php
+++ b/gs-product-configurator/gspc-add-order-meta-to-gf-entry.php
@@ -1,0 +1,127 @@
+<?php
+/**
+ * Gravity Shop // GS Product Configurator // Add order metadata to GF entries on export.
+ *
+ * Instructions:
+ *   1. Install per https://gravitywiz.com/documentation/how-do-i-install-a-snippet/
+ *   2. Update the returned value in `gspc_add_order_meta_to_gf_entry_metas()` accordingly.
+ */
+
+if ( is_callable( 'gs_product_configurator' ) ) :
+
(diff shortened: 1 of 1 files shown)


gs-product-configurator/gspc-add-order-meta-to-gf-entry.php
```diff
@@ -0,0 +1,128 @@
...
+		 *  '_cart_discount',
+		 *  '_created_via',
+		 * );
+		 *
+		 * Please see class WC_Order_Data_Store_CPT in the WC plugin.
+		 * You may also check your DB records or the extension files
+		 * to see other order metadata keys that your installed WC
+		 * extensions might be storing.
+		 */
+		return array();
```
claygriffiths, line 36: Here's a suggestion of moving the example into the returned keys. I think it's safe to put a few defaults here so it will do things simply by adding the snippet.

```suggestion
		 *
		 * Please see class WC_Order_Data_Store_CPT in the WC plugin.
		 * You may also check your DB records or the extension files
		 * to see other order metadata keys that your installed WC
		 * extensions might be storing.
		 */
		return array(
		    '_order_key',
		    '_billing_first_name',
		    '_billing_last_name',
		    '_billing_email',
		    '_billing_phone' => 'Phone',
		    '_shipping_first_name',
		    '_shipping_last_name',
		    '_cart_discount',
		    '_created_via',
		 * );
```

claygriffiths (approved):
LGTM pending suggestions.

PR #806, opened 2024-04-15
`gw-cache-buster.php`: Fixed an issue where entry is marked as spam if honeypot is enabled

barthc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2565804398/64632


spivurno:
Fixed in GF 2.9.

PR #807, opened 2024-04-17
`gpeb-filter-with-default-value.php`: Added snippet to load Entries Block with a provided default value on a filter.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2565332084/64616?folderId=3808239

## Summary

Setup default value for a filter in the Filters block.

Applied the `wp_safe_redirect` logic as on core we do when we manually "apply" a filter ([here](https://github.com/gravitywiz/gp-entry-blocks/blob/master/includes/blocks/class-filters.php#L805-L808))

### DIFF
--- a/gp-entry-blocks/gpeb-filter-with-default-value.php
+++ b/gp-entry-blocks/gpeb-filter-with-default-value.php
@@ -1,0 +1,19 @@
+<?php
+/**
+ * Gravity Perks // Entry Blocks // Apply Default Value to Filter on Page Load
+ * https://gravitywiz.com/documentation/gravity-forms-entry-blocks/
+ *
+ * Filter entries on Page Load with a provided default value
+ */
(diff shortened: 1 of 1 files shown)


claygriffiths (changes requested):
I like the scrappiness of a redirect here, but it's pretty jank.

My inclination is to introduce a new method in GPEB that handles `rgget( 'filters' )` and wraps it with a filter hook.

https://github.com/gravitywiz/gp-entry-blocks/blob/master/includes/class-gf-queryer.php#L363

You'll also likely want to update any `$_GET['filters']` in `class-filters.php` for the block itself so the Filters form is populated.

saifsultanc:
Moved to the plugin https://github.com/gravitywiz/gp-entry-blocks/pull/82

PR #809, opened 2024-04-18
`gpeb-enable-hidden-field-filters.php`: Added snippet to enable Hidden Fields to be viewable as filters.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2570532853/64931?folderId=7098280

## Summary

Adding a filter based on a Hidden field sets the fitler to hidden (it is not visible on the front end)

**Exanoke Filter Settings (Single Line Text B has been marked with Hidden Field Visibility):**
Screenshot 2024-04-18 at 7 31 25 AM

**BEFORE:**
Screenshot 2024-04-18 at 7 29 56 AM

**AFTER:**
Screenshot 2024-04-18 at 7 30 06 AM

### DIFF
--- a/gp-entry-blocks/gpeb-enable-hidden-field-filters.php
+++ b/gp-entry-blocks/gpeb-enable-hidden-field-filters.php
@@ -1,0 +1,24 @@
+<?php
+/**
+ * Gravity Perks // Entry Blocks // Enable Hidden Field Filters
+ * https://gravitywiz.com/documentation/gravity-forms-entry-blocks/
+ *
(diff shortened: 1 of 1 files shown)


claygriffiths (commented):
This feels like it should go into core GPEB.

We have the following method that converts inputs to text: https://github.com/gravitywiz/gp-entry-blocks/blob/master/includes/blocks/class-filters.php#L464-L502

I'd look into why this method isn't working in this case and update it to handle visibility.

If someone is selecting one of these fields to filter by in the Filters block, it's a safe assumption they want it to be visible. 😃

PR #810, opened 2024-04-19
`gw-random-fields.php`: Fixed fields re-appearing in subsequent pages in a multi-page form.

lanresmith:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2567849650/64749/

## Summary

In a multi-page form, when the order of fields is randomized, on subsequent pages, fields may re-appear. Also, fields that are yet to be filled at all might even be re-ordered to previous pages! This PR ensures that for a multi-page form, fields are shuffled _only_ once. This appears to be how the UX can be fixed.

### DIFF
--- a/gravity-forms/gw-random-fields.php
+++ b/gravity-forms/gw-random-fields.php
@@ -21,7 +21,8 @@ class GFRandomFields {
 
 	public $all_random_field_ids;
 	public $display_count;
-	public $selected_field_ids = array();
+	public $selected_field_ids  = array();
+	public $multipage_field_ids = array();
 	public $preserve_order;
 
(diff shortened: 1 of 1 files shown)


spivurno:
Dang. It's clear you've invested some time in this @lanresmith. There's an open PR that has an improved version of this snippet that accounts for some other scenarios. I'm curious if you could try that branch to see if solves this issue:

https://github.com/gravitywiz/snippet-library/pull/334

Let me know if it does not and I'll go ahead and set up a project to roll this entire snippet into [GP Randomizer](https://gravitywiz.com/documentation/gravity-forms-randomizer/). Given your new product knowledge, I'll let you take lead on that. 🤘

PR #812, opened 2024-04-25
`gpaa-show-address-in-language.php`: Added snippet to show address in a specific language in the address fields.

lanresmith:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2577549960/65279

## Summary

When users enter an address, it is populated from Google based on the browser's language. This PR provides a way to force Google to use a chosen language instead.


claygriffiths (approved):
Clean!

S&M with:

```
`gpaa-show-address-in-language.php`: Added snippet to return autocomplete results in a specific language rather than the visitor's language.
```

PR #816, opened 2024-04-29
`gw-multiple-entries-list-field.php`: Added the ability to process feeds per list data row.

lanresmith:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2569827939/64882

## Summary

With this change, feeds are processed per list data row.

### DIFF
--- a/gravity-forms/gw-multiple-entries-list-field.php
+++ b/gravity-forms/gw-multiple-entries-list-field.php
@@ -3,6 +3,8 @@
  * Gravity Wiz // Gravity Forms // Multiple Entries by List Field
  * https://gravitywiz.com/
  *
+ * Instruction Video: https://www.loom.com/share/e253325af6d24cefa20cfd2bdb44fb61
+ *
  * Create multiple entries based on the rows of a List field. All other field data will be duplicated for each entry.
  * List field inputs are mapped to Admin-only fields on the form.
  *
@@ -10,8 +12,15 @@
  * Plugin URI:   https://gravitywiz.com/
  * Description:  Create multiple by entries based on the rows of a List field.
  * Author:       Gravity Wiz
- * Version:      0.7
+ * Version:      0.8
  * Author URI:   https://gravitywiz.com/
+ *
+ * Usage:
+ *
+ * 1. With `append_list_data` set to false (the default), the first row of the list data is actually appended - counter-intuitive, uh? - to the main entry via the admin-only fields and not stored as a separate entry. Subsequent rows are always stored as separate entries.
+ * 2. Feeds are processed per list data row when `process_feeds` is `true`.
+ * 3. Use the admin-only fields - NOT the original List field - in the field maps for feeds.
+ * 4. The original List field data is also stored in the 'child' entries when `preserve_list_data` is `true`.
  */
 class GW_Multiple_Entries_List_Field {
 
@@ -28,6 +37,7 @@ class GW_Multiple_Entries_List_Field {
 				return $value;
 			},
 			'send_notifications' => false,
+			'process_feeds'      => false,
 		) );
 
 		// do version check in the init to make sure if GF is going to be loaded, it is already loaded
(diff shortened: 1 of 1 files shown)


gravity-forms/gw-multiple-entries-list-field.php
```diff
@@ -76,32 +85,33 @@ public function create_multiple_entries( $entry ) {
...
-
 				/**
 				 * Sync the parent entry with our working entry so when it is passed onto other plugins using this filter,
 				 * it is up-to-date and if the entry is updated via this filter (looking at you, GFPaymentAddOn::entry_post_save()),
 				 * our changes will be preserved.
 				 */
-				$entry                          = $working_entry;
-				$entry['gwmelf_parent_entry']   = true;
-				$entry['gwmelf_group_entry_id'] = $working_entry['id'];
+				$entry = $working_entry;
```
claygriffiths, line 97: Does the removal of...

```
$entry['gwmelf_parent_entry']   = true;
$entry['gwmelf_group_entry_id'] = $working_entry['id'];
```

... still work with the `GFPaymentAddOn::entry_post_save()` scenario?

spivurno (commented):
Lookin' good, @lanresmith! A few questions and one suggestion. 😄

gravity-forms/gw-multiple-entries-list-field.php
```diff
@@ -110,9 +120,46 @@ public function create_multiple_entries( $entry ) {
...
+		// Update the passed entry for other filters
+		// that it may be passed to afterwards.
+		$entry['gwmelf_parent_entry']   = true;
+		$entry['gwmelf_group_entry_id'] = $entry['id'];
+
+		if ( $this->_args['append_list_data'] && $this->_args['process_feeds'] ) {
+			/**
+			 * 1st row of list data is added to the DB as a separate entry
+			 * so feed shouldn't be processed for the parent, else we get
+			 * e.g. (for GPGS) an extra row with empty list column values.
```
claygriffiths, line 135: Thanks for the inline comment!

Per the WordPress Coding Standards, we prefer line breaks after around 80 characters for the sake of consistency: https://developer.wordpress.org/coding-standards/inline-documentation-standards/php/#:~:text=or%20%E2%80%9Clink%20tag%E2%80%9D.-,Line%20wrapping,-DocBlock%20text%20should

gravity-forms/gw-multiple-entries-list-field.php
lanresmith: Well, firstly, handling payments is a limitation - even before the changes in the PR - I'll soon add that to the snippet. E.g.: Just one list row is sent to Google Sheets for example, if the form is submitted with 2 list rows & needs payment, else if no payment is needed, the 2 rows are successfully added to GS.

Secondly, the code was not totally removed but moved out of the `foreach` since it should always run, and all it does is just to put some meta in the DB that identifies the 'parent' entry & also update the `$entry` object passed to the callback (since it's a filter) so that other callbacks that receive the entry afterwards & that might want to do something like `GFAPI::update_entry( $entry )` don't overwrite what (i.e. delete the new metadata) we just stored in the DB.

In fact, the former position of the code was preventing having the 'parent' entry properly marked (in the DB) in some cases - that was the issue David also noticed in [the video](https://www.loom.com/share/a4c9fb68ed0342908fcc58ede8e98ad5) (around 1:20).

gravity-forms/gw-multiple-entries-list-field.php
```diff
@@ -42,14 +52,14 @@ public function init() {
 			return;
 		}
 
-		// carry on
-		add_filter( 'gform_entry_post_save', array( $this, 'create_multiple_entries' ) );
+		// We use 5 to ensure that this runs before the ones in GF core.
+		add_filter( 'gform_entry_post_save', array( $this, 'create_multiple_entries' ), 5, 2 );
```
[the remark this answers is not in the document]
lanresmith: This is just to ensure that the callback here runs first. If we use the default priority `10` here, the ones in GF core also have priority `10` and may run 1st and what they do is to process feeds, so feeds might be processed for the 'parent' entry. We might not want that e.g. for the reason stated around lines 131-138:

   ```
 if ( $this->_args['append_list_data'] && $this->_args['process_feeds'] ) {
        .
        .
        .
    }
```

gravity-forms/gw-multiple-entries-list-field.php
lanresmith: Alright, noted. Thanks!

spivurno:
Closing due to complications.

PR #819, opened 2024-04-29
`gpadvs-enable-add-new-option.php`: Added functionality to create new terms on form submission.

barthc:
## Context

⛑️ https://secure.helpscout.net/conversation/2571319811/64967


claygriffiths (commented):
Looking good! I'd add this as a new setting in `$this->_args` such as `insert_new_option` and document how it works and its limitations (e.g. only works with taxonomy terms when used with GPPA)

Also, there are a few PHPCS warnings/errors

claygriffiths (approved):
I like it!

PR #821, opened 2024-05-01
`gppa-acf-repeater-mapper.php`: Added support for custom values with ACF field property template.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2576524151/65219/

## Summary

When using the [GPPA ACF Repeater Mapper](https://github.com/gravitywiz/snippet-library/blob/4bfeb339090e89416c52ea20464a1ff872ae7729/gp-populate-anything/gppa-acf-repeater-mapper.php) snippet, static values set in the Choice Template are ignored.


claygriffiths (approved):
LGTM pending fixing up PHPCS issues

S&M using your existing PR title.

PR #823, opened 2024-05-02
`gw-disable-submit-button-until-required-fields-are-filled-out.php`: Fixed an issue with Disable Submit Button snippet.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2570453216/64916/

## Summary

Disable the submit button until all required fields are filled. The snippet was causing functional issues, as well as syntatic issues. All those have been fixed. 

If the user intent is to work with this using GSPC, then instead of tapping submit button with this `var submitButton = $( '#gform_submit_button_' + self.formId )`, tap the add to cart button instead `var submitButton = $( '.single_add_to_cart_button' )`.

### DIFF
--- a/gravity-forms/gw-disable-submit-button-until-required-fields-are-filled-out.php
+++ b/gravity-forms/gw-disable-submit-button-until-required-fields-are-filled-out.php
@@ -4,7 +4,7 @@
 *
 * Disable submit buttones until all required fields have been filled out. Currently only supports single-page forms.
 *
-* @version   1.0
+* @version   1.1
 * @author    David Smith <david@gravitywiz.com>
 * @license   GPL-2.0+
 * @link      http://gravitywiz.com/...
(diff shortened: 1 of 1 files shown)


gravity-forms/gw-disable-submit-button-until-required-fields-are-filled-out.php
```diff
@@ -63,7 +71,8 @@ public function script() {
 					self.runCheck = function() {
 
 						var submitButton = $( '#gform_submit_button_' + self.formId );
-
+						// For using this snippet with GSPC, remove above line and replace with the line below.
+						// var submitButton = $( '.single_add_to_cart_button' );
```
claygriffiths, line 75: In GPPA, this is the logic we use for finding the submit button. Might help eliminate the need for customizing for GSPC:

```
$form
		.find('.gform_footer, .gform_page_footer, .gfield--type-submit')
		.find('input[type="submit"], input[type="button"]');
```

gravity-forms/gw-disable-submit-button-until-required-fields-are-filled-out.php
```diff
@@ -14,15 +14,23 @@ class GW_Disable_Submit {
...
+		$this->form_id = $form_id;
+
+		add_action( 'gform_pre_render', array( $this, 'maybe_output_script' ) );
+		add_action( 'gform_register_init_scripts', array( $this, 'add_init_script' ) );
 
 	}
 
 	public function maybe_output_script( $form ) {
 
+		if ( $form['id'] !== $this->form_id ) {
```
claygriffiths, line 30:
```suggestion
		if ( $form['id'] != $this->form_id ) {
```

Typically, in snippets we generally stray away from eqeqeq so they're a bit more forgiving in terms of configuration.

claygriffiths (commented):
I have some suggestions. Can you also get the PHPCS stuff resolved?

gravity-forms/gw-disable-submit-button-until-required-fields-are-filled-out.php
saifsultanc: Add to Cart isn't the gform_footer. Also, it's `button[type="submit"]. Tested and updated! Thanks.

claygriffiths (approved):
LGTM!

PR #829, opened 2024-05-10
`gw-require-unique-values.php`: Added support to required unique field inputs.

barthc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2583004576/65519

## Summary

Added support to required unique inputs for `name` field type. Original request was specifying the  inputs id,(not sure if the user can specify the input ids correctly as they are not visible on the editor) but trying to keep it simple by requiring unique value for all inputs.

### DIFF
--- a/gravity-forms/gw-require-unique-values.php
+++ b/gravity-forms/gw-require-unique-values.php
@@ -50,7 +50,6 @@ class GW_Require_Unique_Values {
 		}
 
 		add_filter( sprintf( 'gform_field_validation_%s', $this->_args['form_id'] ), array( $this, 'validate' ), 10, 4 );
-
 	}
 
 	public function validate( $result, $value, $form, $field ) {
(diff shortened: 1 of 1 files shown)


gravity-forms/gw-require-unique-values.php
```diff
@@ -96,6 +95,29 @@ public function validate( $result, $value, $form, $field ) {
...
+			}
+
+			if ( $result['is_valid'] && ! $is_unique ) {
+				$result['is_valid'] = false;
+				$result['message']  = $this->_args['validation_message'];
+			}
+		}
+
+		return $result;
+	}
```
claygriffiths, line 119: I think this method can be removed

claygriffiths (commented):
I think this is still too complicated. The idea is we just want to be able to support input IDs in addition to field IDs no matter the field type.

Just thinking out loud... `get_filtered_value` could be modified to include an optional $input_id param.

Then, `$values[ $field->id ] = $do_hash ? $this->get_value_hash( $value ) : $value;` would need to be updated to have `$field_id` instead of `$field->id`

Lastly, `get_filtered_value` would then use the optional param to get the input value. 

`$field->get_value_submission` should return an array for fields with inputs.

barthc:
```
add_filter( sprintf( 'gform_field_validation_%s', $this->_args['form_id'] ), array( $this, 'validate' ), 10, 4 );
add_filter( sprintf( 'gform_field_validation_%s', $this->_args['form_id'] ), array( $this, 'validate_field_input' ), 11, 4 );
```
@claygriffiths The first filter will support 
`'field_ids' => array( 1.3, 1.6, 2 ), // able to mix inputs and regular fields`
while the second filter will support 
`'field_ids' => array( 1.3, 1.6 ),`

PR #832, opened 2024-05-21
`gspc-validate-date-fields-in-cart.php`: Added validation of date fields in cart.

lanresmith:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2588416647/65817

## Summary

Validation of dates field in WooCommerce cart especially when carts are abandoned (for long), then resumed. E.g.:
On 5/6/24: User adds an entry with date 5/7/24 to their cart but abandons cart.
On 5/8/24: User returns to the site. The entry in the cart now has a date in the past OR a date that's invalid according to GP Limit Dates conditions.


claygriffiths (approved):
LGTM! I'm confirming if we want this as a snippet or in GSPC.

PR #833, opened 2024-05-23
`gw-field-to-field-conditional-logic.php`: Added support for non-numeric rule field ID.

barthc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2600589764/66401

## Summary
The snippet isn't working with GP Inventory's "Available" conditional logic option. The issue is non-numeric ids are excluded from the dependencies array. This PR should fix the issue.


claygriffiths (approved):
LGTM!

PR #834, opened 2024-05-24
`gw-all-fields-template.php`: Added `format` modifier to have the value type choice of `text` or `html`.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2605590008/66639?folderId=7098280

## Summary

When using [All Fields Template](https://gravitywiz.com/gravity-forms-all-fields-template/), the email fields output as a link. 

Screenshot 2024-05-24 at 4 42 11 PM

This is good if we have it on an HTML Block, but not good if we are using it on an Email Field or Text Field.

To fix this, we are adding a `format` modifier which we can set to `text` to override the behaviour.

After the update, if we use `format[text]` with that merge tag `@{Nested Form:1:index[0],filter[3],format[text]}`, we will see:

Screenshot 2024-05-24 at 4 44 07 PM

PS: We have only added the `format` explicitly for the Email field. For the HTML Block, we let it render the default HTML format.

### DIFF
--- a/gravity-forms/gw-all-fields-template.php
+++ b/gravity-forms/gw-all-fields-template.php
@@ -8,7 +8,7 @@
  * Plugin URI:   https://gravitywiz.com/gravity-forms-all-fields-template/
  * Description:  Modify the {all_fields} merge tag output via a template file.
  * Author:       Gravity Wiz
- * Version:      0.10
+ * Version:      0.11
  * Author URI:   http://gravitywiz.com
  *
  * Usage:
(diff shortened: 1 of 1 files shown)


saifsultanc:
@claygriffiths We had discussed this one previously. We had a ticket I had only added `$format = $field->type === 'email' ? 'text' : $format;` check to force the email field value to render as `text` always. We never updated the `format` modifier then, but it was in our discussion. So adding that update now.

claygriffiths:
@saifsultanc I like it, but I am a bit weary of just passing `$raw_value` back. Here's something I did in just GPNF that allows passing the format and it'll change the format used for `GFCommon::get_lead_field_display()`.

[gp-nested-forms-3-08-43 PM.patch](https://github.com/gravitywiz/snippet-library/files/15475517/gp-nested-forms-3-08-43.PM.patch)

saifsultanc:
Closing in favour of https://github.com/gravitywiz/gp-nested-forms/pull/283

PR #835, opened 2024-05-24
`gw-quantity-as-decimal.php`: Added suppport for decimals in feed product quantity.

barthc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2592472572/65993?folderId=8114575

## Summary
When the `GW_Quantity_Decimal`  snippet is used to allow a decimal in the quantity input field, the incorrect price is sent to the payment gateway. This PR would fix the issue by allowing decimals for the feed product quantity.

### DIFF
--- a/gravity-forms/gw-quantity-as-decimal.php
+++ b/gravity-forms/gw-quantity-as-decimal.php
@@ -42,8 +42,11 @@ class GW_Quantity_Decimal {
 
 		if ( $this->global ) {
 			add_filter( 'gform_field_validation', array( $this, 'allow_quantity_float' ), 10, 4 );
+			add_filter( 'gform_submission_data_pre_process_payment', array( $this, 'modify_submission_data' ), 10, 4 );
 		} else {
 			add_filter( 'gform_field_validation_' . $this->form_id, array( $this, 'allow_quantity_float' ), 10, 4 );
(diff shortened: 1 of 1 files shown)


gravity-forms/gw-quantity-as-decimal.php
```diff
@@ -90,6 +93,114 @@ function modify_quantity_input_tag( $markup, $field, $value, $lead_id, $form_id
 		return $markup;
 	}
 
+	function modify_submission_data( $submission_data, $feed, $form, $entry ) {
```
claygriffiths, line 96: This method is pretty hefty, and I'm admittedly having a hard time following what exactly we're doing here.

The appears to be a copy and paste of `get_order_data` with some changes. Is that correct? If so, can you call that out and note what changes you made so if something changes upstream that breaks this snippet, we'll know what to update in the future?

claygriffiths (commented):
Thanks for adding this! I have one request.

@spivurno, I'm also curious what your thoughts are on this. I saw that Barth modified the snippet and sent it as a Gist, but I'm wondering if this particular addition would be better in an experimental snippet instead?

spivurno:
@claygriffiths Given the scope of the change and the burden we're taking on by incorporating so much core code here, I'd be in favor of drafting this (or closing if that is preferred) until we can validate the need.

barthc:
@spivurno  @claygriffiths  I suggest we close this PR.

PR #839, opened 2024-06-06
`gfoai-save-prompt-to-entry.php`: Added the ability to save the final prompt to the entry.

lanresmith:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2612211847/66960

## Summary

Sometimes, a prompt could have a lot of conditionals, so you can't simply copy it using a Live Merge Tag. This snippet grabs the final prompt that's ultimately sent to OpenAI and stores it along with the entry.

### DIFF
--- a/gravityforms-openai/gfoai-save-prompt-to-entry.php
+++ b/gravityforms-openai/gfoai-save-prompt-to-entry.php
@@ -23,7 +23,7 @@ class GFOAI_Save_Prompt_To_Entry {
 			'feed_id'  => false,
 		) );
 
-		$this->_message  = '';
+		$this->_message = '';
 
 		add_filter( 'gf_openai_request_body', array( $this, 'save_request_body' ), 10, 3 );
 		add_action( 'gform_post_process_feed', array( $this, 'process_feed' ), 10, 2 );
--- a/gravityforms-openai/gfoai-save-prompt-to-entry.php
+++ b/gravityforms-openai/gfoai-save-prompt-to-entry.php
@@ -1,0 +1,54 @@
+<?php
+/**
+ * Gravity Wiz // Gravity Forms OpenAI // Save Prompt to Entry
+ *
+ * Sometimes, a prompt could have a lot of conditionals, so you can't simply copy it using a
+ * Live Merge Tag. This snippet grabs the final prompt that's ultimately sent to OpenAI and
+ * stores it along with the entry.
+ *
+ * Instructions:
+ *  1. Install per https://gravitywiz.com/how-do-i-install-a-snippet/
+ */
+
+class GFOAI_Save_Prompt_To_Entry {
+
+	private $_args;
+	private $_message;
+
+	public function __construct( $args = array() ) {
+
+		$this->_args = wp_parse_args( $args, array(
+			'form_id'  => false,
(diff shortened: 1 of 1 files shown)


gravityforms-openai/gfoai-save-prompt-to-entry.php
```diff
@@ -0,0 +1,29 @@
...
+$field_id = '';
+
+add_filter( 'gf_openai_request_body', function ( $body, $endpoint, $feed ) use ( $form_id, $field_id ) {
+	if ( ! $field_id || ( $form_id && $feed['form_id'] != $form_id ) ) {
+		return $body;
+	}
+
+	add_action( 'gform_post_process_feed', function ( $_feed, $entry ) use ( $body, $feed, $field_id ) {
+		if ( $_feed['id'] == $feed['id'] ) {
+			gform_add_meta( $entry['id'], $field_id, $body['messages'][0]['content'] );
```
claygriffiths, line 24: Is the intention to overwrite the actual field content?

If so, it might be worth changing...

```
This snippet grabs the final prompt that's ultimately sent to OpenAI and stores it along with the entry.
```

to...

```
This snippet grabs the final prompt that's ultimately sent to OpenAI and stores it in a field's value.
```

claygriffiths (commented):
One quick question

gravityforms-openai/gfoai-save-prompt-to-entry.php
lanresmith: @claygriffiths I don't think the intention is to overwrite the actual field content. `$field_id` is in fact expected to be an admin (or hidden) field (so it's not filled by the submitter).

[Ticket opener](https://secure.helpscout.net/conversation/2612211847/66960/#thread-7858382856):

> log it into a hidden administrative field

spivurno: @lanresmith Sorry for the confusion, this should be saved to a field value (not to the entry meta). The user wants easy access to the data and getting it out of the entry meta can be a little trickier than fetching it from a field value. The second part of his quoted request refers to a field on the form.

gravityforms-openai/gfoai-save-prompt-to-entry.php
```diff
@@ -0,0 +1,29 @@
...
+$field_id = '4';
+
+add_filter( 'gf_openai_request_body', function ( $body, $endpoint, $feed ) use ( $form_id, $field_id ) {
+	if ( ! $field_id || ( $form_id && $feed['form_id'] != $form_id ) ) {
+		return $body;
+	}
+
+	add_action( 'gform_post_process_feed', function ( $_feed, $entry ) use ( $body, $feed, $field_id ) {
+		if ( $_feed['id'] == $feed['id'] ) {
+			$entry[ $field_id ] = $body['messages'][0]['content'];
```
claygriffiths, line 24: How does this work? Is `$entry` passed as a reference?

saifsultanc: sorry about that. Updated. Please see now!

claygriffiths (approved):
LGTM outside of one suggestion

PR #841, opened 2024-06-10
`gw-advanced-merge-tags.php`: Added `:selected` modifier to target selected checkbox by index on the Advanced Merge Tags snippet.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2616973134/67202

## Summary

Added support for a :selected modifier that would let you target the selected checkbox by its zero-based index in our [Advanced Merge Tags](https://gravitywiz.com/snippet-library/gw-advanced-merge-tags/) snippet.

**How this update works:**
https://www.loom.com/share/0cc9030f5e4a40b59911ff6af3315078

### DIFF
--- a/gravity-forms/gw-advanced-merge-tags.php
+++ b/gravity-forms/gw-advanced-merge-tags.php
@@ -452,12 +452,13 @@ class GW_Advanced_Merge_Tags {
 	 */
 	public function handle_field_modifiers( $value, $input_id, $modifier, $field, $raw_value, $format ) {
 
-		$modifiers = $field->get_modifiers();
+		$modifiers = $this->parse_modifiers( $modifier );
+
 		if ( empty( $modifiers ) ) {
 			return $value;
 		}
 
-		foreach ( $modifiers as $modifier ) {
+		foreach ( $modifiers as $modifier => $modifier_options ) {
 			switch ( $modifier ) {
 				case 'wordcount':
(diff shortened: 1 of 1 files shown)


gravity-forms/gw-advanced-merge-tags.php
```diff
@@ -490,6 +490,13 @@ public function handle_field_modifiers( $value, $input_id, $modifier, $field, $r
 					// Example {My Address Field:1.6:abbr}
 					$default_countries = array_flip( GF_Fields::get( 'address' )->get_default_countries() );
 					return rgar( $default_countries, $value );
+				default:
```
claygriffiths, line 493: Can we use a `case` here and then `parse_modifiers` (or from `$modifiers` if it has the value) instead of the additional regex?

claygriffiths (commented):
This is rad! One suggestion.

gravity-forms/gw-advanced-merge-tags.php
saifsultanc: `$modifier` has the value `selected[0]` or `selected[1]`. We cannot directly add a `case` condition, until we pre-prcocess.

claygriffiths (approved):
LGTM! S&M with:

```
`gw-advanced-merge-tags.php`: Added `:selected` modifier to target selected checkbox by index.
```

PR #845, opened 2024-06-13
`gw-validate-that-a-value-exists.php`: Fixed an issue with validate value not working with AJAX.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2622834298/67455

## Summary

We were erroneously mapping target field's ID when validating on AJAX side. Fixed it with `sourceFieldId`. This fix had nothing to do with GPUID.

Walkthrough of the issue:
https://www.loom.com/share/42e2401ba9854f64acaf52e8970954fb

### DIFF
--- a/gravity-forms/gw-validate-that-a-value-exists.php
+++ b/gravity-forms/gw-validate-that-a-value-exists.php
@@ -222,7 +222,7 @@ class GW_Value_Exists_Validation {
 									break;
 								}
 							}
-							values[ inputId ] = $( this ).val();
+							values[ sourceFieldId ] = $( this ).val();
 						} );
 						return values;
 					}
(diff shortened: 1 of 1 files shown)


gravity-forms/gw-validate-that-a-value-exists.php
```diff
@@ -222,7 +222,7 @@ public function output_script() {
 									break;
 								}
 							}
-							values[ inputId ] = $( this ).val();
+							values[ sourceFieldId ] = $( this ).val();
```
claygriffiths, line 225: Is there a situation in which `sourceFieldId` could be undefined?

I'm wondering if it should be something like...

```suggestion
							values[ sourceFieldId ? sourceFieldId : inputId ] = $( this ).val();
```

claygriffiths (commented):
One question/suggestion.

S&M with:

```
`gw-validate-that-a-value-exists.php`: Fixed an issue with value validation over AJAX.
```

PR #847, opened 2024-06-20
`gpld-block-date-range.php`: Fixed an issue where the end date is not included.

barthc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2629654339/67747?folderId=8114575

## Summary
The snippet doesn't include the End Date in the block dates. This PR will fix the issue.


claygriffiths (approved):
LGTM!

PR #849, opened 2024-06-24
`gw-cache-buster.php`: Fixed issue with Cache Buster not clearing cookies with Paypal checkout and Nested entry.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2630007391/67768/

## Summary

When Cache Buster is enabled, Nested Form fields are not being cleared when paying for a form using PayPal.

_Loom demonstrating the issue:_ https://www.loom.com/share/4816b6c7728a416083cfc706b28ff54a?sid=540ba7fe-133d-4e13-b4fb-b00912b4a181

When we submit form with Stripe checkout, `GPNF_Session::get_session_path` gets the correct session path, however, the same on paypal isn't true. It gets an empty session path. `GPNF_Session::delete_cookie`(L180) relies on getting correct cookie to delete it. Since, the session path isn't correct with paypal it fails. This only happens with Cache Buster.

_xDebug snapshot with Stripe:_
stripe

_xDebug snapshot with Paypal:_
paypal

The fix proposed here is to ensure we have removed all cookies for that form in cache buster scenario.

### DIFF
--- a/gravity-forms/gw-cache-buster.php
+++ b/gravity-forms/gw-cache-buster.php
@@ -39,7 +39,6 @@ class GW_Cache_Buster {
 			return;
 		}
 
-		add_filter( 'gform_pre_render', array( $this, 'clear_gpnf_cookies' ) );
 		add_filter( 'gform_shortcode_form', array( $this, 'shortcode' ), 10, 3 );
 		add_filter( 'gform_save_and_continue_resume_url', array( $this, 'filter_resume_link' ), 15, 4 );
 		add_filter( 'gform_pre_replace_merge_tags', array( $this, 'replace_embed_url' ), 10, 4 );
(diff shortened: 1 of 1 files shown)


claygriffiths (changes requested):
This would nuke cookies for GPNF every time you load a form that is being cached-busted.

Can you do a Git bisect on GPNF to narrow down what the cause is if it used to work?

saifsultanc:
> This would nuke cookies for GPNF every time you load a form that is being cached-busted.
> 
> Can you do a Git bisect on GPNF to narrow down what the cause is if it used to work?

The cause is this commit: https://github.com/gravitywiz/gp-nested-forms/commit/89f3f8ec54618ea8f163e5d07cd6d9a5e25d37be

where we added `get_session_path` in the GPNF Session implementation. It used to work normally before this version

So Cache Buster with this form setup worked fine till GPNF 1.1.34, and broke with v1.1.35.

@claygriffiths

PR #851, opened 2024-06-28
`gw-select-all-multi-select.js`: Added support for dynamically populated Multi Select Fields.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2637029038/68103

## Summary

The snippet above doesn't work when the results of multi select field is dynamically filtered based on a field value.

Loom of the issue (Credits: Samuel)
https://www.loom.com/share/3b402f06f9ff4e02b55ee0b8c5f1a0f2?sid=51bb0ba0-5b42-4c38-84f1-7deaf8c4db31

### DIFF
--- a/gravity-forms/gw-select-all-multi-select.js
+++ b/gravity-forms/gw-select-all-multi-select.js
@@ -14,23 +14,33 @@
 // Update "4" to your Multi Select field ID.
 let targetFieldId = 4;
 
-let $selectAll = $( '<button type="button" style="margin-top: 1rem;">Select All</button>' );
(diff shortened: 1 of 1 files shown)


claygriffiths (approved):
LGTM! It looks like the customer wants to support multiple `targetFieldId`'s. Might be best to change that to an array

saifsultanc:
> LGTM! It looks like the customer wants to support multiple `targetFieldId`'s. Might be best to change that to an array

Gave the customer a solution for that.

PR #859, opened 2024-07-15
`gw-populate-date.php`: Fixed an issue with default time not loading after conditional logic.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2632728563/67897

## Summary

When using the [Populate Date/Time snippet](https://github.com/gravitywiz/snippet-library/blob/master/gravity-forms/gw-populate-date.php), if a Time field is shown with conditional logic, it is not populated.

We need to send the Time field, like the Date field, in a string data format and not array.

Summary of the snippet in action:
https://www.loom.com/share/b19cd15445664b8bb00c9898e0fdbdff

### DIFF
--- a/gravity-forms/gw-populate-date.php
+++ b/gravity-forms/gw-populate-date.php
@@ -244,7 +244,7 @@ class GW_Populate_Date {
 					$hour = 12;
 				}
 			}
-			$date = array( $hour, $minute, $ampm );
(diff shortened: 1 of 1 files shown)


claygriffiths (approved):
LGTM! Any reason behind this? What lead you to this solution? Might be worth putting your findings in an inline comment as this is news to me 😅

PR #860, opened 2024-07-15
`gw-advanced-merge-tags.php`: Fixed an issue with unknown advanced merge tag label.

barthc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2649918005/68702

## Summary

When the `Advanced Merge Tags` snippet is used, the snippet would replace unknown merge tag label with an empty string hence the issue. This PR will add a check for unknown labels.


claygriffiths (approved):
LGTM!

PR #861, opened 2024-07-16
`gpcc-conditional-copy.js`: Added snippet for conditional copy based on trigger field value.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2637685533/68121

## Summary

Copying with condition feature of Copy cat to copy values from one field to another based on if another field contains a value, with extended check to see if the value in the condition field is a specific value.

(version 1):
https://www.loom.com/share/f8db53c9359c45d8bc160d6e1d8eb4fc

**(version 2):**
https://www.loom.com/share/0b02559975444a139519b6830c7fc6d8

### DIFF
--- a/gp-copy-cat/gpcc-conditional-copy.js
+++ b/gp-copy-cat/gpcc-conditional-copy.js
@@ -11,12 +11,30 @@
 *
 * 2. Configure the snippet per the inline instructions.
 */
-gform.addFilter( 'gpcc_copied_value', function( value, $targetElem, field ) {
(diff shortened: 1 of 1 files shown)


saifsultanc:
Snippet updated as per more elaborate customer requirements, and updated loom: https://www.loom.com/share/0b02559975444a139519b6830c7fc6d8

claygriffiths (approved):
Good stuff!

PR #866, opened 2024-07-29
`gpnf-hide-expired-nested-form-field.php`: Added snippet to Hide Expired Nested Form Field.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2660146653/69147

## Summary

Snippet to hide a Nested Form field if the child form has expired (due to a schedule being set up on the child form). Currently, if the child form has expired, the modal will still open and display the form, but there will be an error when adding the entry:

**Demo:**
https://www.loom.com/share/798f30d73d254864aa6a65e7d16d390b

### DIFF
--- a/gp-nested-forms/gpnf-hide-expired-nested-form-field.php
+++ b/gp-nested-forms/gpnf-hide-expired-nested-form-field.php
@@ -12,7 +12,7 @@ function check_nested_form_expiration( $form ) {
 		return $form;
 	}
 
-	foreach( $form['fields'] as &$field ) {
+	foreach ( $form['fields'] as &$field ) {
 		if ( $field->type == 'form' && $field->gpnfForm ) {
 			$nested_form = GFAPI::get_form( $field->gpnfForm );
 			$is_expired  = GFFormDisplay::validate_form_schedule( $nested_form );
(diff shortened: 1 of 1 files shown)


gp-nested-forms/gpnf-hide-expired-nested-form-field.php
```diff
@@ -0,0 +1,25 @@
+<?php
+/**
+ * Gravity Perks // GP Nested Forms // Hide Nested Form Field if the Nested Form has expired.
+ * https://gravitywiz.com/documentation/gravity-forms-nested-forms/
+ *
+ * Instruction Video: https://www.loom.com/share/798f30d73d254864aa6a65e7d16d390b
+ */
+add_filter( 'gform_pre_render', 'check_nested_form_expiration' );
+function check_nested_form_expiration( $form ) {
```
claygriffiths, line 9: If you keep this as a function and not a class, I'd either prefix it or go with an anonymous function.

gp-nested-forms/gpnf-hide-expired-nested-form-field.php
```diff
@@ -0,0 +1,25 @@
...
+	// Update '435' to the targeted Parent Form's ID.
+	if ( rgar( $form, 'id' ) != 435 || ! is_callable( 'gp_nested_forms' ) ) {
```
claygriffiths, line 11: Thoughts on making this accept an array? Might also be worth making it so you can pass some sort of value that makes it global.

claygriffiths (approved):
I like it. Couple of suggestions.

PR #870, opened 2024-08-08
`gpnf-child-products-in-parent-order-summary.php`: Fixed an issue where nested form product field total in not include in woocommerce cart total.

barthc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2670405760/69619

## Summary

When the [Include Child Products Directly in Parent Form Order Summary](https://github.com/gravitywiz/snippet-library/blob/master/gp-nested-forms/gpnf-child-products-in-parent-order-summary.php) snippet is active on a site that uses GSPC, the total from the Nested Forms is not added to the Woo product.

The fix here is to modify the child product key within the loop.

### DIFF
--- a/gp-nested-forms/gpnf-child-products-in-parent-order-summary.php
+++ b/gp-nested-forms/gpnf-child-products-in-parent-order-summary.php
@@ -50,7 +50,7 @@ add_filter( 'gform_product_info', function( $product_info, $form, $entry ) {
 						}
 					}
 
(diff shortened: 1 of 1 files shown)


gp-nested-forms/gpnf-child-products-in-parent-order-summary.php
```diff
@@ -1,3 +1,4 @@
+
```
claygriffiths, line 1:
```suggestion
```

claygriffiths (approved):
LGTM! Can you S&M with what you have after you remove the added whitespace?

PR #871, opened 2024-08-15
`gspc-sync-field-max-to-wc-quantity.js`: Fixed an issue with max attribute when input is mapped as WooCommerce quantity.

barthc:
## Context
When the input is mapped as woocommerce quantity, the max input attribute is not set to the maximum quantity of the selected product. This PR fixed the issue by using the `found_variation` action to listen to changes instead of using the mutation observer.

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2677576845/69900


claygriffiths (approved):
It's always a great PR when a `MutationObserver` is removed.

LGTM!

PR #872, opened 2024-08-17
`gpld-inline-datepicker-only-for-mobile-devies.php`: Added snippet for inline datepicker to work only on mobile devices.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2675073655/69822

## Summary

Totally missed creating this PR. Doing so now..

Snippet to only apply inline date picker for mobile devices.

EDIT: Lost track of this totally, from what i recall the [JS hook](https://docs.gravityforms.com/gform_datepicker_options_pre_init/) suggested by Dave was causing issues with the broken display on the non-inline datepicker (desktop view)


claygriffiths (approved):
LGTM, I like the simplicity of it.

If you remember your findings with a JS approach that would be more caching friendly, can you document them in this PR?

PR #873, opened 2024-08-20
`gpnf-copy-parent-value-manually.js`: Fixed an issue where source field value is not copied over if it is a dropdown or radio button field.

barthc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2670546355/69628

## Summary
The `gpnf-copy-parent-value-manually.js` snippet doesn't work when the parent form source field is a Radio button field.
This PR would fix the issue.

### DIFF
--- a/gp-nested-forms/gpnf-copy-parent-value-manually.js
+++ b/gp-nested-forms/gpnf-copy-parent-value-manually.js
@@ -25,7 +25,9 @@ gform.addAction( 'gpnf_init_nested_form', function( childFormId, gpnf ) {
 
 function copyParentFormValue( parentFormId, parentFieldId, childFormId, childFieldId ) {
 
-	var value = jQuery ( '#input_{0}_{1}'.gformFormat( parentFormId, parentFieldId ) ).val();
+	var value = jQuery ( '#input_{0}_{1}'.gformFormat( parentFormId, parentFieldId ) ).val() ||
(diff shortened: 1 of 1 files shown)


gp-nested-forms/gpnf-copy-parent-value-manually.js
```diff
@@ -25,7 +25,9 @@ gform.addAction( 'gpnf_init_nested_form', function( childFormId, gpnf ) {
 
 function copyParentFormValue( parentFormId, parentFieldId, childFormId, childFieldId ) {
 
-	var value = jQuery ( '#input_{0}_{1}'.gformFormat( parentFormId, parentFieldId ) ).val();
+	var value = jQuery ( '#input_{0}_{1}'.gformFormat( parentFormId, parentFieldId ) ).val() ||
+		jQuery('input[name="input_{0}"]:checked'.gformFormat( parentFieldId )).val() ||
+		jQuery('input[name="input_{0}"]:selected'.gformFormat( parentFieldId )).val();
```
claygriffiths, line 30: Just stating this for posterity, this whole `copyParentFormValue` is very basic and by the looks of it, would likely only work for copying a single value into a text/textarea input. I'm not sure how complicated we want to make this function.

claygriffiths (approved):

PR #874, opened 2024-08-21
`gpnf-disable-sessions.php`: Fixed an issue where non privileged users are not able to edit or delete entries.

barthc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2677531768/69897

## Summary
When the GPNF disable session snippet is active, non-privileged users can not edit or delete entries. This PR will fix the issue.


spivurno:
@barthc Closing this PR in favor of #877. This approach felt a little clunky by requiring the user to specify both a parent and child form ID in the code and presented an opportunity for malicious users to populate existing child entries into a Nested Form field for which they should have no access.

PR #876, opened 2024-08-23
`gw-require-unique-values.php`: Fixed issue with Product Value comparison.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2686989633/70358

## Summary

The snippet doesn't work when the fields involved are Product Choice fields having similar label/value property but different prices.

### DIFF
--- a/gravity-forms/gw-require-unique-values.php
+++ b/gravity-forms/gw-require-unique-values.php
@@ -184,6 +184,10 @@ class GW_Require_Unique_Values {
 			$value = basename( rgars( $_FILES, sprintf( 'input_%d/name', $field->id ) ) );
 		} else {
 			$value = $field->get_value_submission( array() );
+			// Product values are stored as Value|Price, we just need to compare Value.
+			if ( $field->type == 'product' && strpos( $value, '|' ) ) {
+				$value = explode( '|', $value )[0];
+			}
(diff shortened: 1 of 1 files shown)


gravity-forms/gw-require-unique-values.php
```diff
@@ -184,6 +184,10 @@ public function get_filtered_value( $field, $input_id = null ) {
 			$value = basename( rgars( $_FILES, sprintf( 'input_%d/name', $field->id ) ) );
 		} else {
 			$value = $field->get_value_submission( array() );
+			// Product values are stored as Value|Price, we just need to compare Value.
+			if ( $field->type == 'product' && strpos( $value, '|' ) ) {
+				$value = explode( '|', $value )[0];
+			}
```
claygriffiths, line 190:
```suggestion
			// Product values are stored as Value|Price, we just need to compare Value.
			if ( rgar( $field, 'enablePrice' ) ) {
				$value = GFCommon::get_selection_value( $value );
			}
```

claygriffiths (commented):
Quick suggestion—this is how I see GF doing this check.

claygriffiths (approved):
LGTM, S&M

PR #880, opened 2024-09-03
`gpep-edit-entry.php`: Fixed an issue where cached product info are used for total field calculation.

barthc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2692325023/70631

## Summary
The GPEPT Edit Entry snippet doesn't update the total product fields on the backend.  The issue is that the cached product info is used to run the total calculation. The fix here is to purge the cache before the calculation.

### DIFF
--- a/gp-easy-passthrough/gpep-edit-entry.php
+++ b/gp-easy-passthrough/gpep-edit-entry.php
@@ -97,6 +97,9 @@ class GPEP_Edit_Entry {
 
 		$update_entry_id = $this->get_edit_entry_id( $form['id'] );
 		if ( $update_entry_id ) {
+			// Purge product info cache
+			$this->purge_product_cache( $form, GFAPI::get_entry( $update_entry_id ) );
+
 			if ( $this->delete_partial
 				&& is_callable( array( 'GF_Partial_Entries', 'get_instance' ) )
 				&& $entry_id !== null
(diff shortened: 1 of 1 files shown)


gp-easy-passthrough/gpep-edit-entry.php
```diff
@@ -246,6 +249,24 @@ public function exclude_edit_entry_from_inventory( $query, $field ) {
...
+		);
+
+		foreach ( $cache_options as $cache_option ) {
+			list( $use_choice_text, $use_admin_label ) = $cache_option;
+			if ( gform_get_meta( rgar( $entry, 'id' ), "gform_product_info_{$use_choice_text}_{$use_admin_label}" ) ) {
+				gform_delete_meta( rgar( $entry, 'id' ), "gform_product_info_{$use_choice_text}_{$use_admin_label}" );
+			}
+		}
+
+	}
```
claygriffiths, line 268: This looks super similar `GFFormsModel::refresh_product_cache()`. What led you to duplicate it rather calling it directly?

claygriffiths (approved):
LGTM pending one question.

S&M with:

```
`gpep-edit-entry.php`: Fixed an issue where cached product could be used when editing which could result in improper calculations and totals.
```

gp-easy-passthrough/gpep-edit-entry.php
barthc: @claygriffiths  the `GFFormsModel::refresh_product_cache()` is calling the `GFCommon::get_product_fields` [here](https://github.com/gravityforms/gravityforms/blob/master/forms_model.php#L3504) and the issue persist but the updated value will show up on further edit.

claygriffiths: Got it! Thank you

PR #884, opened 2024-09-07
`gpld-conditional-limits.php`: Added a new snippet to enable conditional date limits.

barthc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2690682399/70534

## Summary

Added a new snippet to enable conditional date limits.

### DIFF
--- a/gp-limit-dates/gpld-conditional-limits.php
+++ b/gp-limit-dates/gpld-conditional-limits.php
@@ -1,0 +1,203 @@
+<?php
+/**
+ * Gravity Wiz // Gravity Perks // GP Limit Dates // Conditional Limits
+ * https://gravitywiz.com/documentation/gravity-forms-limit-dates/
+ *
+ * Provide conditional date options for your date fields.
+ *
+ */
+class GPLD_Conditional_Limits {
+
+	private $_args = array();
+
+	public function __construct( $args = array() ) {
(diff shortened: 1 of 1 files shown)


spivurno (changes requested):
Dude!! Very excited about this. Some feedback:

https://www.loom.com/share/279be099b3024bc6a9b1b04701300eb3

1. Set settings by default so they don’t have to manually mapped if they’re unaffected.
    - Also, some properties are not supported (e.g. `inlineDatepicker`).
2. Need to output script in the footer. See: https://github.com/gravitywiz/snippet-library/blob/master/gw-snippet-template.php#L83C2-L91C3
3. Need to enqueue `conditional_logic.js` if this script is used.

spivurno:
@barthc Getting close!

A few lingering issues:
https://www.loom.com/share/504667b49e644a35876c3156188ea573

- [ ] Let's clear the date when configuration changes.
- [ ] "Invalid Date" warning does not show up when conditional configuration is applied.
- [ ] Invalid dates are not triggering validation errors when conditional configuration is applied.

spivurno (approved):
Working well! 

One note: It's preferable to just make multiple commits during the review process rather than force pushing changes. It allows me to see what changed which makes code review a lot easier on subsequent reviews. 🙏

PR #891, opened 2024-09-19
`gw-none-of-the-above-checkbox.js`: Added setting to not disable None of the Above by default.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2711484484/71608

## Summary

Variable added at the top for specifying whether the "None of the Above" should be disabled - default to false which will change the default behavior of this snippet.

Loom Demonstration:
https://www.loom.com/share/a142a6a2a4da481b83d1d18937c12747

### DIFF
--- a/gravity-forms/gw-none-of-the-above-checkbox.js
+++ b/gravity-forms/gw-none-of-the-above-checkbox.js
@@ -21,9 +21,10 @@
  */
 $( '.gw-none-of-the-above' ).each( function() {
 
-	var $field  = $( this );
-	var $last   = $field.find( '.gchoice:last-child input' );
-	var $others = $field.find( 'input' ).not( $last );
+	var $disable_nota = false;
+	var $field        = $( this );
+	var $last         = $field.find( '.gchoice:last-child input' );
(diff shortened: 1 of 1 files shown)


gravity-forms/gw-none-of-the-above-checkbox.js
```diff
@@ -21,9 +21,10 @@
  */
 $( '.gw-none-of-the-above' ).each( function() {
 
-	var $field  = $( this );
-	var $last   = $field.find( '.gchoice:last-child input' );
-	var $others = $field.find( 'input' ).not( $last );
+	var $disable_nota = false;
```
claygriffiths, line 24:
```suggestion
	var disable_nota = false;
```

This isn't a jQuery object so we wouldn't usually prefix this with a `$`

claygriffiths (approved):
LGTM pending variable name change.

S&M with:

```
`gw-none-of-the-above-checkbox.js`: Added variable to control whether "None of the Above" gets disabled. It now defaults to not being disabled.
```

PR #894, opened 2024-09-24
`gspc-add-entry-data-to-woo-order-custom-field.php`: Added new snippet.

matty0501:
## Context

⛑️ Ticket(s): [](https://secure.helpscout.net/conversation/2685171079/70219/)

## Summary

This snippet will update a custom field on the Woo order with data from the attached GF entry, when the order is created.

For example, a customer collects a name in the GF entry, and needs to send this to ShipStation with the Woo order. ShipStation expects custom data to be stored in a specific custom field on the Woo order. This snippet will allow you to add the data from the entry into that custom field on the order.


spivurno (approved):

spivurno:
Approved, @matty0501!

Some notes about the PHPCS and Danger warnings.

https://www.loom.com/share/63936f7581f34a02b756e1db122039b2

PR #897, opened 2024-09-26
`gw-progress-meter.php`: Fixed an issue where refunds are included in the total paid count.

barthc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2711207909/71589

## Summary
When the total paid count is calculated refunds are included, this PR will the issue by excluding refunds.

### DIFF
--- a/gravity-forms/gw-progress-meter.php
+++ b/gravity-forms/gw-progress-meter.php
@@ -100,7 +100,8 @@ class GW_Progress_Meter {
 					'join'   => '',
 					'where'  => $wpdb->prepare( "
 						WHERE e.form_id = %d
-						AND e.status = 'active'\n",
+						AND e.status = 'active'
+						AND e.payment_status = 'Paid'\n",
 						$atts['id']
 					),
 				);


gravity-forms/gw-progress-meter.php
```diff
@@ -114,7 +115,8 @@ public function get_count( $atts ) {
 					'where'  => $wpdb->prepare( "
 						WHERE em.form_id = %d
 						AND em.meta_key = %s
-						AND e.status = 'active'\n",
+						AND e.status = 'active'
+						AND e.payment_status = 'Paid'\n",
```
claygriffiths, line 119: @spivurno Should we be adding this conditional in both of these spots? Thinking out loud, what if the form doesn't have payments wired up?

As an aside, should the partial entries bit be moved outside of the else to right above `$sql`?

spivurno: @claygriffiths We'll only want this in the first condition that checks for the presence of the payment_amount. This entry property is only populated if a payment feed is configured (regardless of the presence of product fields on the form).

As for partial entries, I don't think it's possible for a payment to be captured on a partial entry... maybe with Gravity Flow? My gut is to not change that until we have a bug report.

claygriffiths: > We'll only want this in the first condition that checks for the presence of the payment_amount.

Thought so but wasn't 100% sure 😄

> As for partial entries, I don't think it's possible for a payment to be captured on a partial entry... maybe with Gravity Flow? My gut is to not change that until we have a bug report.

Makes sense!

claygriffiths (approved):

PR #898, opened 2024-09-27
`gppa-lmt-extract-first-character.php`: Added new snippet.

matty0501:
## Context

⛑️ Ticket(s): [](https://secure.helpscout.net/conversation/2671741596/69678/)

## Summary

This snippet will return the first character of the string when using a Live Merge Tag.

For example, taking a Name field and extracting the first character of the name.


spivurno (approved):
💯

PR #900, opened 2024-09-30
`gpnf-set-created-by-property-on-child-entries-after-user-registration.php`: Added snippet for GP Nested Forms that sets the created by property on the child entries after user account is created.

sbassah:
Added new snippet for GP Nested Forms that sets the created by property on the child entries after user account is created.

## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2719134672/71947/

## Summary

When a nested form is submitted by a non-logged-in user, the created by property is not set on both the parent form entry and the child entries. However, when a, create user feed, is on the parent form to create a user account, it sets the created by property on the parent form but doesn't set that of the child form entries. This snippet sets the created by property on the child form entries, that is linked to the user registration parent form.

### DIFF
--- a/gp-nested-forms/gpnf-set-created-by-property-on-child-entries-after-user-registration.php
+++ b/gp-nested-forms/gpnf-set-created-by-property-on-child-entries-after-user-registration.php
@@ -1,0 +1,27 @@
+<?php
+/**
+ * Gravity Perks // Nested Forms // Set Created By Propery on Child Entries After User Account Registration.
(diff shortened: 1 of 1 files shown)


spivurno:
@sbassah Wanna get the PHP/Danger issues fixed up and we'll get this merged?

sbassah:
@spivurno Sorry, I didn't see the tag earlier.

 Saif has got the PHP/Danger issue sorted out, and it's ready for review.

spivurno (approved):

PR #902, opened 2024-10-03
`gpeb-process-feeds-on-edit.php`: Added the ability to exclude certain feeds from being processed.

barthc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2720318050/72019

## Summary
Added the ability to exclude certain feeds from being processed by specifying their feed IDs.


spivurno:
@barthc A few considerations:

1. To make this easier to configure lets pass the `$excluded_feed_ids` variable to the anonymous function with the `use` keyword. This would allow them to set that variable at the [top of the snippet](https://github.com/gravitywiz/snippet-library/pull/902/files#diff-ca7cbacb6c56e44e0c0bb2f0d866682423d412006e80aa7498c12762f070c6feR12) and not have to go hunting for it down here.
2. Add a comment explaining that they can pass an empty array to not exclude any feeds.
3. Should be be removing this function after processing? It may not be necessary but worth considering.

claygriffiths (approved):

PR #903, opened 2024-10-04
`gpfup-enforce-exact-dimensions.js`: Added snippet to enforce exact dimensions on file upload.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2720536211/72026/

💬 Slack: https://gravitywiz.slack.com/archives/GMP0ZMNSE/p1727779512582999

## Summary

Snippet to force users to upload an image of a specific size. So if a user uploads a different-sized image, a validation error will be thrown.

Current GPFUP code logic only adds validation setting for min dimensions [here](https://github.com/gravitywiz/gp-file-upload-pro/blob/44623cfbe4c8ffaee8352f013baf8e92655df9b0/js/src/frontend/GPFUPField.ts#L522-L578)

We are going to use the filter for that and apply the exact settings.

Loom Demo:
https://www.loom.com/share/4b28e2dbdbbb4b399b7220d3d77d71f5

### DIFF
--- a/gp-file-upload-pro/gpfup-enforce-exact-dimensions.js
+++ b/gp-file-upload-pro/gpfup-enforce-exact-dimensions.js
@@ -1,0 +1,30 @@
+/**
+ * Gravity Perks // File Upload Pro // Enforce Exact Width or Height
+ * https://gravitywiz.com/path/to/article/
+ *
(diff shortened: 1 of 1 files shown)


claygriffiths (approved):
Very nice!

PR #907, opened 2024-10-15
`gpnf-disable-sessions.php`: Fixed an issue where nested form fields are not dynamically populated.

barthc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2721901699/72085

## Summary
The `gpnf-disable-sessions` snippet prevents parameter-based dynamic population from functioning in the child form. The request vars are stored in the session and then used to populate the nested form field. Since this snippet disables sessions, the dynamic population won't work, hence the issue.

The fix here is to store the request vars in a cookie and use it to populate the nested form fields.

### DIFF
--- a/gp-nested-forms/gpnf-disable-sessions.php
+++ b/gp-nested-forms/gpnf-disable-sessions.php
@@ -15,6 +15,10 @@
 add_action( 'wp_ajax_gpnf_session', 'gw_gpnf_disable_session', 9 );
 add_action( 'wp_ajax_nopriv_gpnf_session', 'gw_gpnf_disable_session', 9 );
 function gw_gpnf_disable_session() {
+	// Store request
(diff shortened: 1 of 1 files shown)


spivurno (changes requested):
@barthc This snippet is intended to be a solution for the too-large cookie issue we began solving here:

https://github.com/gravitywiz/gp-nested-forms/pull/276/

I'm hesitant to reintroduce a cookie here. 😅

I think the play might be to continue the work of simplifying the cookie by moving the query parameters from the cookie and pass them to the AJAX requests as we did with the field values.

https://github.com/gravitywiz/gp-nested-forms/pull/276/files#diff-39903e76cb5b16955fb5a0619493700912ebd6255c73220522f358a79de23577R2262

PR #910, opened 2024-10-25
`gw-disable-submit-button-until-required-fields-are-filled-out.php`: Fixed a compatibility issue with GP Populate Anything.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2740666744/72953

## Summary

The Disable Submit Button Until Required Fields Are Filled Out snippet is conflicting with GPPA, causing the form to break. Populated fields just pulse and don't update.

Matt A's loom explaining the issue:
https://www.loom.com/share/e98e2369db5a43278f5d2706612fad54

After the update (fix):
https://www.loom.com/share/76f4bf19d9554486bde1a39e74e97ea4

### DIFF
--- a/gravity-forms/gw-disable-submit-button-until-required-fields-are-filled-out.php
+++ b/gravity-forms/gw-disable-submit-button-until-required-fields-are-filled-out.php
@@ -32,14 +32,15 @@ class GW_Disable_Submit {
 		}
 
 		if ( ! self::$script_output ) {
-			$this->script();
+			add_action( 'wp_footer', array( $this, 'output_script' ) );
+			add_action( 'gform_preview_footer', array( $this, 'output_script' ) );
 			self::$script_output = true;
 		}
 
 		return $form;
 	}
 
-	public function script() {
+	public function output_script() {
 		?>
 
 		<script type="text/javascript">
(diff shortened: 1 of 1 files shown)


gravity-forms/gw-disable-submit-button-until-required-fields-are-filled-out.php
```diff
@@ -32,14 +32,15 @@ public function maybe_output_script( $form ) {
 		}
 
 		if ( ! self::$script_output ) {
-			$this->script();
+			add_action( 'wp_footer', array( $this, 'output_script' ) );
+			add_action( 'gform_preview_footer', array( $this, 'output_script' ) );
```
claygriffiths, line 36: Was this to prevent AJAX errors?

Could we use `gform_print_scripts` instead?

saifsultanc: Yes, and I will check `gform_print_scripts`

saifsultanc: Doesnt work with `gform_print_scripts`, i think its because it not firing at the correct time.

claygriffiths: Works for me considering this is a snippet. Thanks for checking!

claygriffiths (approved):

PR #916, opened 2024-11-05
`gpls-use-entry-for-each-user-rule-message.php`: Added new snippet to use the current user entry to replace message merge tag variables.

barthc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2730567828/72487

## Summary

When the rule group is set `User` , this snippet will use the current user's latest entry to process the replace merge variable for the limit message.


claygriffiths (approved):
@barthc Code LGTM and PR title LGTM.

Can you fix the PHPCS issues then S&M this?

PR #921, opened 2024-11-19
`gcn-add-page-icon.php`: Added example to add/update an emoji page icon to a Notion page.

veryspry:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2762766200/73911?viewId=6987275 

## Summary

Example snippet to show how to set a Notion page icon when adding and updating pages.

Requires the changes in  https://github.com/gravitywiz/gc-notion/pull/19 to function but is backwards compatible (e.g. will not break if using an older GCN version).


gc-notion/gcn-add-page-icon.php
```diff
@@ -0,0 +1,29 @@
+<?php
+/**
+ * Gravity Connect // Notion // Add Page Icon
```
veryspry, line 3: @veryspry update minimum GCN version once the related GCN changes are released.

claygriffiths (approved):

PR #927, opened 2024-11-26
`gw-disable-submit-button-until-required-fields-are-filled-out.php`: Fixed compatibility issue with image choice field.

barthc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2769979131/74318

## Summary

Fixed compatibility issue with image choice field.

### DIFF
--- a/gravity-forms/gw-disable-submit-button-until-required-fields-are-filled-out.php
+++ b/gravity-forms/gw-disable-submit-button-until-required-fields-are-filled-out.php
@@ -47,6 +47,21 @@ class GW_Disable_Submit {
 
 			var GWDisableSubmit;
 
+			function gwDisableSubmitDebounce(func, wait, immediate) {
+				var timeout;
+				return function() {
+					var context = this, args = arguments;
+					var later = function() {
+						timeout = null;
+						if (!immediate) func.apply(context, args);
+					};
+					var callNow = immediate && !timeout;
+					clearTimeout(timeout);
+					timeout = setTimeout(later, wait);
+					if (callNow) func.apply(context, args);
+				};
+			};
+
 			(function($){
 
 				GWDisableSubmit = function( args ) {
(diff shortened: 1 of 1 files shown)


gravity-forms/gw-disable-submit-button-until-required-fields-are-filled-out.php
```diff
@@ -88,6 +88,12 @@ public function output_script() {
 						$(document).on( 'gppa_updated_batch_fields', function() {
 							self.runCheck();
 						});
+						// Check when field is image choice.
+						gform.addAction( 'gform_input_change', function( elem, formId, fieldId ) {
+							if ( args.inputHtmlIds.includes( `#input_${formId}_${fieldId}` )  &&  $( elem ).parents( '.gfield--type-image_choice' ) ) {
```
claygriffiths, line 93: I'd just put `self.runCheck()` in this action to cover more bases. No need for Image Choice check IMO.

If performance becomes an issue, you could add a super simple debounce around `runCheck` such as:

```js
function gwDisableSubmitDebounce(func, wait, immediate) {
	var timeout;
	return function() {
		var context = this, args = arguments;
		var later = function() {
			timeout = null;
			if (!immediate) func.apply(context, args);
		};
		var callNow = immediate && !timeout;
		clearTimeout(timeout);
		timeout = setTimeout(later, wait);
		if (callNow) func.apply(context, args);
	};
};
```

barthc: What do we now do with the `onchange` event? Keep or remove it?
```
$( args.inputHtmlIds.join( ', ' ) ).change( function() {
      self.runCheck();
} );
```

claygriffiths: Keep it for backwards compatibility and on the off chance that `gform_input_change` doesn't fire.

A debounce of like 5-10ms would result in virtually zero noticeable delay, while preventing the logic from running multiple times when an input value changes.

claygriffiths:
@barthc Thanks for making the changes!

I went ahead and opened https://github.com/gravityforms/gravityforms/issues/3050, so this PR will be put on hold

PR #928, opened 2024-12-02
`gca-create-relation.php`: Added snippete demonstrating how to create a relation with GCA.

veryspry:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2768092049/74169?viewId=6987275 

## Summary

Adds a snippet to demonstrate how to create a Relation in Airtable while processing a GC Airtable Feed.

Relies on: https://github.com/gravitywiz/gc-airtable/pull/4

### DIFF
--- a/gc-airtable/gca-create-relation.php
+++ b/gc-airtable/gca-create-relation.php
@@ -1,0 +1,55 @@
+<?php
+/**
+ * Gravity Connect // Airtable // Create Phone Number Relation
+ *
+ * This snippet demonstrates how to create a relation between two tables in Airtable
+ * when a GC Airtable feed is being processed. It uses an example assuming the following:
+ *
+ * 1. There are a "People" table and "Phone Numbers" table in Airtable.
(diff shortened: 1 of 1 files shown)


claygriffiths (approved):
I'd denote this as an experimental snippet, since it requires a specific configuration.

Before making this non-experimental, I'd convert it to a [class-based snippet](https://github.com/gravitywiz/snippet-library/blob/master/gw-snippet-template.php) which would help make configuration easier and allow for more relationships.

veryspry:
> I'd denote this as an experimental snippet, since it requires a specific configuration.
> 
> Before making this non-experimental, I'd convert it to a [class-based snippet](https://github.com/gravitywiz/snippet-library/blob/master/gw-snippet-template.php) which would help make configuration easier and allow for more relationships.

Sounds good!

I updated the header here https://github.com/gravitywiz/snippet-library/pull/928/commits/cbe0d9b236646c9c9d924c1d06bc0f4aa1bb0c89

PR #933, opened 2024-12-04
`gpnf-auto-attach-parent-files.php`: Added snippet to auto attach parent files to child notifications.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2765560391/74024

## Summary

A snippet to take a file that has been uploaded to the parent form, and attach it to the child form notification.

The snippet we already have which does the reverse (i.e.  attach child form files to parent form notification): https://github.com/gravitywiz/snippet-library/blob/master/gp-nested-forms/gpnf-auto-attach-child-files.php

Demo of this new snippet -

https://www.loom.com/share/36d23d10f9214ba1aa53ed08d214dddc


claygriffiths (approved):
LGTM pending fixing up PHPCS issue

PR #934, opened 2024-12-05
`gpnf-detach-child-entry-from-parent.php`: Added snippet to display a metabox on Entry details for nested entries to detach from Parent Entry.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2779018255/74896

## Summary

Snippet to add a metabox to child entry detail views that, when clicked, will detach the child entry from the parent entry. 

Demo:
https://www.loom.com/share/84a6fc86c75449baa544cd8080c3ed15


claygriffiths (approved):
Amazing work! Code and Loom both look great.

S&M

PR #948, opened 2024-12-13
`gpuid-fill-sequence-gaps.php`: Fixed an issue with the snippet not working.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2790631747/75478

## Summary

The snippet does not seem to work, it hasn't been update in over 2 years. The DB we are accessing is `wp_rg_lead_detail`, but it seems GF 2.3+ it's expected that the `wp_gf_entry_meta` should be used instead. GF codebase for context: https://github.com/gravityforms/gravityforms/blob/master/export.php#L892-L910

So we just apply the DB update, and that fixes the issue.

Here is how it works after the update:
https://www.loom.com/share/1819b282338d4832a68ce985787b16bb

### DIFF
--- a/gp-unique-id/gpuid-fill-sequence-gaps.php
+++ b/gp-unique-id/gpuid-fill-sequence-gaps.php
@@ -13,15 +13,15 @@ add_filter( 'gpui_sequential_unique_id_pre_insert_519_5', function ( $uid, $form
 	global $wpdb;
 
 	$result = $wpdb->get_results( $wpdb->prepare( "
-		select value 
-		from {$wpdb->prefix}rg_lead_detail 
-		where form_id = %d and field_number = %d", $form_id, $field_id ) );
+		select meta_value 
+		from {$wpdb->prefix}gf_entry_meta 
+		where form_id = %d and meta_key = %d", $form_id, $field_id ) );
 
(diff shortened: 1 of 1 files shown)


gp-unique-id/gpuid-fill-sequence-gaps.php
```diff
@@ -13,15 +13,15 @@
 	global $wpdb;
 
 	$result = $wpdb->get_results( $wpdb->prepare( "
-		select value 
-		from {$wpdb->prefix}rg_lead_detail 
-		where form_id = %d and field_number = %d", $form_id, $field_id ) );
+		select meta_value 
+		from {$wpdb->prefix}gf_entry_meta
```
claygriffiths, line 17: I'd follow what RG does here

```
$table_name  = GFFormsModel::get_entry_meta_table_name();
```

claygriffiths (approved):
One quick suggestion regarding how to get the table name.

Otherwise, LGTM. This will increase the minimum version requirement to 2.3, but I think we can all agree that's fine for a snippet 😃

When ready, S&M with:

```
`gpuid-fill-sequence-gaps.php`: Updated to use database schema in Gravity Forms 2.3 and newer.
```

PR #978, opened 2025-01-03
`gpadvs-enable-add-new-option.php`: Added ability to add new options when "No Results" are returned from Populate Anything.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2806641061/76034

## Summary

The functionality to Add New with the snippet above doesn't work if no result is populated into the field.

BEFORE (Samuel):
https://www.loom.com/share/fc8ada4515724f3e8a76294da9910c33?sid=c12dbaf0-3a19-4265-b7a5-4435ac2f8e71

AFTER:
https://www.loom.com/share/fc15e97fed40415899d55be5fc84339e


claygriffiths (approved):
I like it. The only potential suggestion I have is to make the “No Results” text more easily customizable using this snippet.

PR #1014, opened 2025-01-29
`gw-delay-feed-till-user-registration.php`: Added snippet to delay a feed until the User Registration feed had its user activated.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2826984822/77005

## Summary

The user had a registration form that has a GF User Registration feed and Notion feed. They want to send the user ID of the account that is created when the form is submitted to Notion. Currently, it appears the notion feed is processed before the account is created, so it doesn't get the user ID. They are looking for a way to delay the Notion feed until the account is created, and the user ID is assigned to the entry created by property.

The snippet added makes it extensible to other combination with the User Registration feed, by just specifying a different addon slug (like gc-openai or any other).

Instruction Video:
https://www.loom.com/share/176456848c614ab6a3792ca1f9ed9c14

### DIFF
--- a/gravity-forms/gw-delay-feed-till-user-registration.php
+++ b/gravity-forms/gw-delay-feed-till-user-registration.php
@@ -7,7 +7,7 @@
  * Delay a specified feed until user registration is complete.
  * This is useful when you have a feed that needs to be processed after user registration is complete.
  */
-class GW_Delayed_Feed_Processing {
+class GW_Delayed_Feed_Processing_Till_User_Registration {
 
 	private $_args = array();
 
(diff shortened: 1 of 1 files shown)


gravity-forms/gw-delay-feed-till-user-registration.php
```diff
@@ -0,0 +1,72 @@
+<?php
+/**
+ * Gravity Wiz // Gravity Forms // Delay a feed until User Registration is Complete.
+ *
+ * Instruction Video: https://www.loom.com/share/176456848c614ab6a3792ca1f9ed9c14
+ *
+ * Delay a specified feed until user registration is complete.
+ * This is useful when you have a feed that needs to be processed after user registration is complete.
+ */
+class GW_Delayed_Feed_Processing {
```
claygriffiths, line 10: I'd update this to more closely match the filename.

claygriffiths (approved):
Very nice! One quick suggestion. After that, S&M

gravity-forms/gw-delay-feed-till-user-registration.php
saifsultanc: Done

PR #1023, opened 2025-02-18
`gppa-use-acf-choice-label.php`: Fixed PHP fatal error when ACF field set return value as return `Both (array)`.

malayladu:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2846332119/77984

## Summary

User is attempting to grab the Label of an ACF field using this snippet: https://gravitywiz.com/snippet-library/gppa-use-acf-choice-label/.

This works as intended when the ACF field's return value is set to `label`, but their field is set to `Both (Array)`, so the following error is being thrown.

`Uncaught Error: str_contains(): Argument #1 ($haystack) must be of type string, array given
in /wp-includes/blocks.php on line 1817`

This PR fixes this fatal error.

### DIFF
--- a/gp-populate-anything/gppa-use-acf-choice-label.php
+++ b/gp-populate-anything/gppa-use-acf-choice-label.php
@@ -22,5 +22,10 @@ add_filter( 'gppa_process_template_value', function( $template_value, $field, $t
 
 	$label = get_field( str_replace( 'meta_', '', $field->{'gppa-values-templates'}['value'] ), $object->ID );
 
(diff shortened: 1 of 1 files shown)


saifsultanc (approved):
Looks to be working well now! Good job.

I just wonder if we could add some extensibility to the snippet to make it work for "Both" scenarios in "Both" ways as well. Maybe have a different CSS class for this, and then return the both data as `label : value`.

For now this is good, but just a thought. 

Marking approved. Final 👀  from @claygriffiths

claygriffiths (approved):

PR #1024, opened 2025-02-19
`gw-update-posts.php`: Fixed issue where snippet returned ID instead of value when GPPA dynamically populates post field.

malayladu:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2844608881/77895

## Summary

By default, when we use GPPA to get a term, like category of a post, it returns the Term name. However, when this snippet  which is used to update posts, is also active on the website, Populate Anything will return the Term ID when getting the category of a post instead of the name. The set up is what I showed in the video.

So the expected behaviour is that, even with the snippet active on the website, Populate Anything should still return the Term Name when populating Terms of a post.

Here's a video demo of the issue 

https://www.loom.com/share/afe25a1453834809aa249fc342d52205?sid=dc46a634-5708-4627-b91c-789a6109a4ce

### DIFF
--- a/gravity-forms/gw-update-posts.php
+++ b/gravity-forms/gw-update-posts.php
@@ -279,6 +279,16 @@ class GW_Update_Posts {
 	 * @return mixed
 	 */
 	public function return_ids_instead_of_names( $value, $field, $template_name, $populate, $object, $object_type, $objects, $template ) {
+		// Check if this is for the specific form we want.
+		if ( $field->formId != $this->_args['form_id'] ) {
+			return $value;
+		}
+
+		// Don't want to return IDs for post objects used in the populates field dynamically using GPPA.
+		if ( rgar( $field, 'gppa-values-enabled' ) === true  && rgar( $field, 'gppa-values-object-type' ) === 'post' ) {
+			return $value;
+		}
+
 		if ( strpos( $template, 'taxonomy_' ) === 0 ) {
 			$taxonomy = preg_replace( '/^taxonomy_/', '', $template );
 			$terms    = wp_get_post_terms( $object->ID, $taxonomy, array( 'fields' => 'ids' ) );


gravity-forms/gw-update-posts.php
```diff
@@ -279,6 +279,16 @@ function acf_get_field_object_by_name( $field_name, $group_name = false ) {
 	 * @return mixed
 	 */
 	public function return_ids_instead_of_names( $value, $field, $template_name, $populate, $object, $object_type, $objects, $template ) {
+		// Check if this is for the specific form we want.
+		if ( $field->formId != $this->_args['form_id'] ) {
+			return $value;
+		}
+
+		// Don't want to return IDs for post objects used in the populates field dynamically using GPPA.
```
saifsultanc, line 287: Not needed.

saifsultanc (approved):
I was a little confused over the overall scenario. I think your form id check is gold, and we don't need anything else. Please remove the other gppa check.

gravity-forms/gw-update-posts.php
```diff
@@ -140,27 +140,98 @@ public function update_post_by_entry( $entry, $form ) {
...
+				$terms = explode( ',', is_object( $term_field ) ? trim( $term_field->get_value_export( $entry ) ) : '' );
+
+				if ( ! empty( $terms ) && rgar( $term_field, 'gppa-choices-enabled' ) ) {
+					$taxonomy = '';
+					// Get the taxonomy from the field settings or the first choice object based on the field type.
+					if ( $term_field instanceof GF_Field_Text ) {
+						$taxonomy = str_replace( 'taxonomy_', '', rgars( $term_field, 'gppa-values-templates/value', '' ) );
+					} elseif ( is_object( $term_field ) && ! empty( $term_field['choices'][0]['object']->taxonomy ) ) {
+						$taxonomy = $term_field['choices'][0]['object']->taxonomy;
+					} elseif ( is_object( $term_field ) && ! empty( $term_field['gppa-choices-filter-groups'][0][0]['property'] ) && $term_field['gppa-choices-filter-groups'][0][0]['property'] === 'taxonomy' ) {
```
veryspry, line 158: I like to leverage `rgars()` for stuff like this to make the conditional slightly shorter

```suggestion
					} elseif ( is_object( $term_field ) && rgars( $term_field, 'gppa-choices-filter-groups/0/0/property' ) === 'taxonomy' ) {
```

gravity-forms/gw-update-posts.php
```diff
@@ -140,27 +140,98 @@ public function update_post_by_entry( $entry, $form ) {
...
+					// Taxonomy not found, skip.
+					if ( empty( $taxonomy ) ) {
+						continue;
+					}
+
+					foreach ( $terms as $key => $term ) {
+						// Check if `$term` is a term name or id. If term name, get the term id.
+						if ( ! is_numeric( $term ) ) {
+							$term          = term_exists( $term, $taxonomy );
+							$terms[ $key ] = $term ? $term['term_id'] : 0;
```
veryspry, line 172: Curious, why set this to `0` if no `term_id` is found?

gravity-forms/gw-update-posts.php
```diff
@@ -331,4 +412,4 @@ public function get_post_date( $entry, $form ) {
 		return gmdate( 'Y-m-d H:i:s', strtotime( sprintf( '%s %s:%s:00', $date, $hour, $min ) ) );
 	}
 
-}
+}
```
veryspry, line 415: We typically leave a newline at the end of files. Mostly to help ensure compat with certain older tools and it's also a bit better in case something gets concatenated at the end of the file.

```suggestion
}

```

veryspry: https://stackoverflow.com/questions/2287967/why-is-it-recommended-to-have-empty-line-in-the-end-of-a-source-file

gravity-forms/gw-update-posts.php
veryspry: (This one is an optional suggestion)

veryspry (approved):
@malayladu this is untested by me, but the code looks good. I did add a couple of suggestions and a question. I'm going to go ahead and approve so that you can move this along as soon as you decide what (if anything) to do with suggestions 🙂

gravity-forms/gw-update-posts.php
malayladu: @veryspry Yes, you are right. Why we need to set with 0? Unset it if we don't find it. I've modified the code.

gravity-forms/gw-update-posts.php
malayladu: Done!

malayladu:
@veryspry I've made changes based on the feedback. Ready for review and then S&M!

veryspry (approved):
@malayladu thank you! And also approved

PR #1025, opened 2025-02-21
Update gpaa-single-line-as-input.php with instructions on using multiple address fields

coryarobbins:
Added instructions on how to use the snippet for multiple address fields.

## Summary

The configuration block can be added multiple times to support adding multiple address fields in the same form or over different forms.  There was a question about it in the loom video.

### DIFF
--- a/gp-address-autocomplete/gpaa-single-line-as-input.php
+++ b/gp-address-autocomplete/gpaa-single-line-as-input.php
@@ -132,6 +132,8 @@ class GPAA_Single_Line_Input {
 
 }
 
+# Configuration
+
 // Duplicate then modify the next 7 lines to add another address field
 // Configuration for the first address field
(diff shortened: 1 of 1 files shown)


gp-address-autocomplete/gpaa-single-line-as-input.php
```diff
@@ -130,10 +132,12 @@ public function is_applicable_form( $form ) {
 
 }
 
-// Configuration
+// Duplicate then modify the next 7 lines to add another address field
+// Configuration for the first address field
```
spivurno, line 136:
```suggestion
# Configuration

// Duplicate then modify the next 7 lines to add another address field
// Configuration for the first address field
```

spivurno (approved):
Appreciate the PR, @coryarobbins!

PR #1035, opened 2025-02-28
`gpnf-override-parent-merge-tag-on-submission.php`: Added support to rename uploaded files in child entries.

malayladu:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2849364535/78123

Related PR: https://github.com/gravitywiz/gp-file-renamer/pull/22

## Summary

The customer has a nested form with an Upload field. They are using the `{Parent}` merge tag in a field, and then using this in GPFR to name the uploaded file, based on a value in the parent form.

However, they want it so if the user changes the value on the parent form before submitting, the uploaded file is renamed.

### DIFF
--- a/gp-nested-forms/gpnf-override-parent-merge-tag-on-submission.php
+++ b/gp-nested-forms/gpnf-override-parent-merge-tag-on-submission.php
@@ -122,6 +122,15 @@ class GPNF_Override_Parent_Merge_Tags {
 			$child_entry = GFAPI::get_entry( $child_entry_id );
 			$value       = GFCommon::replace_variables( $default_value, $child_form, $child_entry );
 			GFAPI::update_entry_field( $child_entry_id, $input_id, $value );
+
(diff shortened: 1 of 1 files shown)


gp-nested-forms/gpnf-override-parent-merge-tag-on-submission.php
```diff
@@ -122,6 +122,15 @@ function override_child_entry_input_value( $entry, $field, $child_form, $input_i
 			$child_entry = GFAPI::get_entry( $child_entry_id );
 			$value       = GFCommon::replace_variables( $default_value, $child_form, $child_entry );
 			GFAPI::update_entry_field( $child_entry_id, $input_id, $value );
+
+			// Added support to force the renaming of uploaded files if the GP File Renamer plugin is active and the version is greater than 1.0.9.
```
veryspry, line 126: Can you leave a note about why 1.0.9 specifically? (For our future selves)

veryspry (approved):

PR #1038, opened 2025-03-03
`gw-coupons-exclude-products.php`: Added functionality to exclude discounts on options of a product field.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2856409848/78480

## Summary

Updating the snippet so it is possible to exclude product Option fields from a coupon.

For example, the customer has a Product field with a price of $100, and an Option field (attached to the Product field) with a price of $10. If the user applied a 100% coupon, this should not apply to the Option field, so the form total should be $10.

Loom:
https://www.loom.com/share/1149b9eed4d74afbac8d94c5ae2a9e0a

### DIFF
--- a/gravity-forms/gw-coupons-exclude-products.php
+++ b/gravity-forms/gw-coupons-exclude-products.php
@@ -6,7 +6,7 @@
  *
  * Requires Gravity Forms Coupons v1.1
  *
- * @version 1.3
+ * @version 1.4
  */
 class GW_Coupons_Exclude_Products {
 
@@ -19,10 +19,11 @@ class GW_Coupons_Exclude_Products {
 
 		// set our default arguments, parse against the provided arguments, and store for use throughout the class
 		$this->_args = wp_parse_args( $args, array(
-			'form_id'                => false,
-			'exclude_fields'         => array(),
-			'exclude_fields_by_form' => array(),
-			'skip_for_100_percent'   => false,
+			'form_id'                        => false,
+			'exclude_fields'                 => array(),
+			'exclude_fields_without_options' => array(),
+			'exclude_fields_by_form'         => array(),
+			'skip_for_100_percent'           => false,
 		) );
 
 		// do version check in the init to make sure if GF is going to be loaded, it is already loaded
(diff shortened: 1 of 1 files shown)


gravity-forms/gw-coupons-exclude-products.php
```diff
@@ -102,10 +103,11 @@ function output_script() {
...
 				function getExcludedAmount( formId ) {
 
-					var excludeFields = gf_global.gfcep[ formId ],
-						amount        = 0;
+					var excludeFields               = gf_global.gfcep[ formId ].exclude_fields,
+						excludeFieldsWithoutOptions = gf_global.gfcep[ formId ].exclude_fields_without_options,
+						amount                      = 0;
 
-					if( ! excludeFields ) {
+					if( ! excludeFields && ! excludeFieldsWithoutOptions ) {
```
veryspry, line 110: Should this be `||`?

Or in other words, shouldn't the calculations for `excludeFields` and `excludeFieldsWithoutOptions` be independant and not matter if the other exists or not?

gravity-forms/gw-coupons-exclude-products.php
```diff
@@ -19,10 +19,11 @@ public function __construct( $args ) {
 
 		// set our default arguments, parse against the provided arguments, and store for use throughout the class
 		$this->_args = wp_parse_args( $args, array(
-			'form_id'                => false,
-			'exclude_fields'         => array(),
-			'exclude_fields_by_form' => array(),
-			'skip_for_100_percent'   => false,
+			'form_id'                        => false,
+			'exclude_fields'                 => array(),
+			'exclude_fields_without_options' => array(),
```
veryspry, line 24: Since this is about excluding the _options_ and not the field itself, it might make more sense to call this: (or something similar)

```suggestion
			'exclude_options_from_fields' => array(),
```

gravity-forms/gw-coupons-exclude-products.php
veryspry: Followup: this question can be ignored if `excludeFields` and `excludeFieldsWithoutOptions` default to empty arrays 🙂

gravity-forms/gw-coupons-exclude-products.php
```diff
@@ -170,11 +186,18 @@ function modify_coupon_discount_amount( $discount, $coupon, $price ) {
 			return $discount;
 		}
 
-		$price    = $price - $this->excluded_total;
-		$currency = new RGCurrency( GFCommon::get_currency() );
-		$amount   = $currency->to_number( $coupon['amount'] );
+		$orig_price = $price;
```
veryspry, line 189: `$orig_price` looks to be unused

```suggestion
```

veryspry (changes requested):
Just a few minor change requests 🙂

veryspry (approved):
thanks mate 🙌

PR #1040, opened 2025-03-03
`gpeb-remove-fields-on-edit.php`: Fixed incorrect title in the header.

matty0501:
## Context

Just something I noticed.

## Summary

The snippet was incorrectly titled "Set Fields as Readonly On Edit", this PR renames it to "Remove Fields On Edit".


saifsultanc (approved):
Thank you @matty0501 !

LGTM, just S&M. When you do, please remove the PR number from the end.

EDIT: I thought to do a quick walkthrough of how this would be: https://www.loom.com/share/9082be685ca24d05b349cc5d0553335b. Probably you are aware already, but just mentioning. LMK if you have any question!

PR #1042, opened 2025-03-07
`gw-min-and-max-character-limit.php`: Added support to run validation only when the field has a value.

malayladu:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2867998139/78938

## Summary

Min Max character limit validation make sense when the field has a value. If we want to make sure field has a value, that can be handled through field `Required` rules. 

So, this we can do min max character limit validation only for the filed that has a value.

### DIFF
--- a/gravity-forms/gw-min-and-max-character-limit.php
+++ b/gravity-forms/gw-min-and-max-character-limit.php
@@ -85,6 +85,11 @@ class GW_Minimum_Characters {
 			}
 		}
 
+		// If value is empty, don't do validation.
+		if ( empty( $value ) ) {
+			return $result;
+		}
+
 		$char_count      = mb_strlen( $value );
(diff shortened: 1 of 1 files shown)


saifsultanc (approved):
Working well!

This does the job as David emphasized, "I think we can ignore the required status and only apply our validation if the value is not empty."

However, I do wonder if we can make it more versatile by adding a boolean flag, say:
```php
'ignore_if_empty' => true
```

Maybe a future improvement scenario for the snippet.

For now, LGTM!

spivurno:
@saifsultanc Love where your head is at! 

But I think this is probably best without an additional parameter since this is arguably how it should have been working all along. If this were a plugin, we'd need to be more considerate of backwards compatibility for folks who may have been using this min/max validation (without having marked the field as required) but fortunately, it's a snippet. 😄

saifsultanc:
@malayladu S&M this. Thanks! 🙌

PR #1044, opened 2025-03-08
`gwaft-template-collapsible.php`: Fixed an issue with page numbers and hidden fields.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2849211669/78114

## Summary

The collapsible templates snippet does not work because the values of `$page_groups` and `$pages` are an "index" behind. 

Screenshot 2025-03-08 at 10 11 55 AM

Adjustment for the index issue, and adding a field visibility check.

**BEFORE** Customer described issue (was also reproduced locally with a simple form):
https://app.screencast.com/wVMGdfNrt8gY2

**AFTER** the snippet update:
https://www.loom.com/share/457a4df1ae314b92a54e5f18b3e7ad71

### DIFF
--- a/gravity-forms/gwaft-template-collapsible.php
+++ b/gravity-forms/gwaft-template-collapsible.php
@@ -25,7 +25,14 @@ add_filter( 'gwaft_template_output', function( $content, $slug, $name, $data, $s
 	$pages = $data['form']['pagination']['pages'];
 	$page_groups = array();
 	foreach ( $data['items'] as $item ) {
-		$page_groups[ $item['field']['pageNumber'] ][] = $item;
+		$field = $item['field'];
+		// Skip hidden fields.
+		if ( $field->type == 'hidden' || $field->visibility == 'hidden' ) {
+			continue;
+		}
+		// Adjust pageNumber to be zero-based.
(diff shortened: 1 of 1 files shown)


claygriffiths:
@coderabbitai review

gravity-forms/gwaft-template-collapsible.php
```diff
@@ -25,7 +25,14 @@
 	$pages = $data['form']['pagination']['pages'];
 	$page_groups = array();
 	foreach ( $data['items'] as $item ) {
-		$page_groups[ $item['field']['pageNumber'] ][] = $item;
+		$field = $item['field'];
+		// Skip hidden fields.
+		if ( $field->type == 'hidden' || $field->visibility == 'hidden' ) {
```
veryspry, line 30: I can't really see a situation in which type casting would be useful here 🧐 

```suggestion
		if ( $field->type === 'hidden' || $field->visibility === 'hidden' ) {
```

veryspry (approved):
I left one suggested change, but am approving anyways since I'm OOO the next few days. Don't want to hold a snippet up for a nitpick 🙂

PR #1047, opened 2025-03-13
`gcoai-stream-to-paragraph-field.js`: Fixed duplicate code and an issue with snippet not working.

SebastianWiz:
## Context

⛑️ Ticket(s): [https://secure.helpscout.net/conversation/2853989407/78340?viewId=8172236](https://secure.helpscout.net/conversation/2853989407/78340?viewId=8172236)

## Summary

If the Response field is set to a plain text Paragraph field and there are rich text enabled Paragraph fields elsewhere on the form, the `gcoai-stream-to-paragraph-field` snippet fails to copy OpenAI's output to the Response field. This happens because the code checks if TinyMCE is globally available, not whether the specific Response field has TinyMCE enabled.

This PR fixes this by modifying the TinyMCE detection to specifically check if the target Response field has TinyMCE enabled, rather than just checking if TinyMCE exists globally.

Additionally, this PR removes duplicate code at the end of the snippet.

### DIFF
--- a/gc-openai/gcoai-stream-to-paragraph-field.js
+++ b/gc-openai/gcoai-stream-to-paragraph-field.js
@@ -27,15 +27,15 @@ $streamFieldInput.on( 'change', function() {
 	$input = $( `#input_GFFORMID_${responseFieldId}` );
 	var inputValue = this.value;
 
-	// Get HTML from response field if TinyMCE is available.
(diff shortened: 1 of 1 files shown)


saifsultanc (approved):
LGTM!

Thanks for getting rid of the duplicate code. It was my mistake 😭

PR #1052, opened 2025-03-19
`gwaft-template-collapsible.php`: Added Order Summary as an additional last page.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2872221797/79232

## Summary

When using the [all fields collapsible sections](https://github.com/gravitywiz/snippet-library/blob/master/gravity-forms/gwaft-template-collapsible.php) snippet, if there is a pricing field on the form, the order summary is included in its own section. However, this also causes the page names to be showing incorrectly on the sections, and the section the order summary is included in does not have a title.

Matt A's loom explaining the issue:
https://www.loom.com/share/48c97b4b77054d36ab5c64973bd8ea84

Handle the "Order Summary" page has been proposed with this PR.

How this update works:
https://www.loom.com/share/728fc541438d4041a8e8e91ceee87e58

### DIFF
--- a/gravity-forms/gwaft-template-collapsible.php
+++ b/gravity-forms/gwaft-template-collapsible.php
@@ -29,6 +29,7 @@ add_filter( 'gwaft_template_output', function( $content, $slug, $name, $data, $s
 
 		// Add Order Summary as a separate last page.
 		if ( ! $field && rgar( $item, 'label' ) == apply_filters( 'gwaft_order_summary_label', 'Order Summary' ) ) {
+			// Storing order summary page as '-1' to be the last page, and avoid conflicts with the actual page numbers.
 			$page_groups[-1][] = $item;
 			$pages[-1]         = apply_filters( 'gwaft_order_summary_label', 'Order Summary' );
 			continue;
--- a/gravity-forms/gwaft-template-collapsible.php
+++ b/gravity-forms/gwaft-template-collapsible.php
@@ -25,9 +25,17 @@ add_filter( 'gwaft_template_output', function( $content, $slug, $name, $data, $s
 	$pages = $data['form']['pagination']['pages'];
 	$page_groups = array();
 	foreach ( $data['items'] as $item ) {
-		$field = $item['field'];
+		$field = rgar( $item, 'field' );
+
+		// Add Order Summary as a separate last page.
+		if ( ! $field && rgar( $item, 'label' ) == apply_filters( 'gwaft_order_summary_label', 'Order Summary' ) ) {
+			$page_groups[-1][] = $item;
+			$pages[-1]         = apply_filters( 'gwaft_order_summary_label', 'Order Summary' );
+			continue;
+		}
+
 		// Skip hidden fields.
-		if ( $field->type === 'hidden' || $field->visibility === 'hidden' ) {
+		if ( ! $field || $field->type === 'hidden' || $field->visibility === 'hidden' ) {
 			continue;
 		}
 		// Adjust pageNumber to be zero-based.


gravity-forms/gwaft-template-collapsible.php
```diff
@@ -25,9 +25,17 @@
 	$pages = $data['form']['pagination']['pages'];
 	$page_groups = array();
 	foreach ( $data['items'] as $item ) {
-		$field = $item['field'];
+		$field = rgar( $item, 'field' );
+
+		// Add Order Summary as a separate last page.
+		if ( ! $field && rgar( $item, 'label' ) == apply_filters( 'gwaft_order_summary_label', 'Order Summary' ) ) {
```
veryspry, line 31: This check on the `label` value feels potentially flimsy to me

gravity-forms/gwaft-template-collapsible.php
```diff
@@ -25,9 +25,17 @@
...
+			$page_groups[-1][] = $item;
```
veryspry, line 32: Can you explain how the `-1` indexing works here?

gravity-forms/gwaft-template-collapsible.php
saifsultanc: This is how the `order summary` item is:
Screenshot 2025-03-29 at 4 06 19 PM


For regular fields, it is:
Screenshot 2025-03-29 at 4 06 41 PM

So an alternate logic could possibly be "if ( ! field )". But, that could possibly be a false flag is what I wondered.

gravity-forms/gwaft-template-collapsible.php
saifsultanc: I was storing `-1` as an arbitary and additional page containing the detail for 'Order Summary'. Do you want to do it as 99129 (or the like)? A big arbitrary number instead.

veryspry: Got it!

Would it work to push the order summary  item to the end of `$page_groups` and then set the `$page` index accordingly?

gravity-forms/gwaft-template-collapsible.php
veryspry: That's tricky.

Looking at the code in [gw-all-fields-template.php](https://github.com/gravitywiz/snippet-library/blob/fd45ba6ca714d3e20dbbbc47e51c85f924b80216/gravity-forms/gw-all-fields-template.php), it _looks_ like `$field` will always be present _unless_ it's the product summary.

The product summary is added [here](https://github.com/gravitywiz/snippet-library/blob/fd45ba6ca714d3e20dbbbc47e51c85f924b80216/gravity-forms/gw-all-fields-template.php#L532) and the other two instances (where `$field` is included) are [here](https://github.com/gravitywiz/snippet-library/blob/fd45ba6ca714d3e20dbbbc47e51c85f924b80216/gravity-forms/gw-all-fields-template.php#L445) and [here](https://github.com/gravitywiz/snippet-library/blob/fd45ba6ca714d3e20dbbbc47e51c85f924b80216/gravity-forms/gw-all-fields-template.php#L517)

veryspry: That's tricky.

A second option, is to modify the `gw-all-fields-template` snippet so that the order summary item has an additional key / flag that could be checked here. (Maybe `is_order_summary => true`?).

Given this is a snippet, I'm okay leaving this as is. 

@saifsultanc can you add a comment explaining the conditional / how it's expected to work?

gravity-forms/gwaft-template-collapsible.php
saifsultanc: I was vary of other potential customizations that may be thrown it by customers, adding additional pages and stuff? Thought to keep it 'separate' from the rest of the stuff. Unless if you suggest that pushing it at the end is a safer way to go forth, I will make the update. Thanks!

veryspry: I do think throwing it on the end is a safer bet, but that said, I'm actually not super familiar with other potential customizations etc. I'll leave the final call to you!

If you do go with `-1`, could you add a note explaining the logic?

veryspry (approved):
@saifsultanc thank you! Added one final note, but I'll let you decide the final approach 🙂

PR #1055, opened 2025-03-25
`gppa-populate-child-entries.php`: Fixed an issue with missing cookie values.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2882383665/79731

## Summary

The temporary parent entry ID doesn't populate the field on the form after the form is submitted and opened again. The cookie wasn't found strangely. Added logic to look out for the cookie in this context.

### DIFF
--- a/gp-populate-anything/gppa-populate-child-entries.php
+++ b/gp-populate-anything/gppa-populate-child-entries.php
@@ -66,7 +66,20 @@ class GPPA_Populate_Child_Entries {
 		} else {
 			$session = new GPNF_Session( $field->formId );
 			$cookie  = $session->get_cookie();
-			$value   = rgar( $cookie, 'hash', '' );
+
+			// try to find cookie, if available.
+			if ( ! $cookie ) {
+				$cookie_name = 'gpnf_form_session_';
+				$pattern = '/^' . preg_quote( $cookie_name, '/' ) . '[\w\W]*/';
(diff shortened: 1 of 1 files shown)


gp-populate-anything/gppa-populate-child-entries.php
```diff
@@ -66,7 +66,20 @@ public function populate_parent_entry_id( $value, $field ) {
 		} else {
 			$session = new GPNF_Session( $field->formId );
 			$cookie  = $session->get_cookie();
-			$value   = rgar( $cookie, 'hash', '' );
+
+			// try to find cookie, if available.
+			if ( ! $cookie ) {
```
spivurno, line 71: So a cookie exists and but it's not retrieved by the session? Could you explain more about this scenario and why the cookie is not being found?

saifsultanc: The standard cookie `gpnf_form_session` wasn't there, which I thought was to do with some logic on the snippet. But there was another one with the a suffix of `gfpreview` and form Id (forgot the exact format) available. The proposal here is to look out for that if the standard isn't available.

spivurno: I think we'll want to identify why the expected cookie isn't there and why the expected cookie is there.

PR #1063, opened 2025-04-01
`gppa-lmt-remove-commas.php`: Added new snippet.

matty0501:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2893446866/81553?viewId=14960

## Summary

This snippet removes commas from the copied value when using a Live Merge Tag. The specific use case it was written for is when copying values over 1,000 from a Number field to a Quantity field. The comma was preventing the value from being copied correctly.


saifsultanc (approved):
Lovely. LGTM. 

Just squash & merge (S&M)!

PR #1065, opened 2025-04-01
`gw-user-registration-prevent-populating-mapped-fields.php`: Added a snippet to prevent mapped fields from being populated with saved values from the User Registration feed.

malayladu:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2875771201/79417

## Summary

This snippet prevents mapped fields from being populated with saved values from the User Registration feed.

### DIFF
--- a/gravity-forms/gw-user-registration-prevent-populating-mapped-fields.php
+++ b/gravity-forms/gw-user-registration-prevent-populating-mapped-fields.php
@@ -1,0 +1,26 @@
+<?php
+/**
+ * Gravity Wiz // Gravity Forms User Registration // Prevent Population Of Mapped Fields With The Saved Values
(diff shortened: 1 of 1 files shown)


saifsultanc (approved):
This looks good. You could maybe add an "instructions" block, minimalistic like this can also be handy for future customers : https://github.com/gravitywiz/snippet-library/blob/master/gravity-forms/display-number-of-entries-left.php#L8-L12

Overall, LGTM!

malayladu:
@saifsultanc I've added instructions block. Should I do S&M?

saifsultanc:
> @saifsultanc I've added instructions block. Should I do S&M?

Yes!

PR #1066, opened 2025-04-04
`gpnf-limit-entry-min-max-from-field.php`: Fixed an issue with the snippet not working correctly on form load.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2897379455/81715

## Summary

When using the [Dynamically Set Entry Min/Max From Field Value](https://gravitywiz.com/snippet-library/gpnf-limit-entry-min-max-from-field/) snippet, the Nested Form shows "maximum number reached" before any child entry is even added.

The key issue was default-ing to a zero max value.


veryspry (approved):
I like it 🙂‍↕️

PR #1071, opened 2025-04-10
`gw-rich-text-html-fields.php`: Added functionality to remember the last selected `Visual` or `HTML` editor mode.

malayladu:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2901552081/81917

## Summary

After saving in Text mode, the editor reverts back to Visual mode. This PR added functionality to remember the last selection.


spivurno:
@malayladu Not a review, just a hot thought – my instinct would be to save this to the field in the form meta. This way if they open the field in another browser they'll still have the same preference set.

malayladu:
> @malayladu Not a review, just a hot thought – my instinct would be to save this to the field in the form meta. This way if they open the field in another browser they'll still have the same preference set.

That’s a great idea for keeping the choice persistent across devices and browsers. However, to make that work, we’d also need to save user-specific preferences for the form. If multiple people are working on the same form, we’ll need to store preferences individually for each user.

Do we want to implement this?

saifsultanc:
> However, to make that work, we’d also need to save user-specific preferences for the form. If multiple people are working on the same form, we’ll need to store preferences individually for each user.
> 
> Do we want to implement this?


You are right. Since form meta applies globally, we'd need to introduce a way to store these preferences on a per user basis. We could potentially explore storing the preference using user meta key to the form ID?

saifsultanc (approved):

spivurno:
@malayladu @saifsultanc @claygriffiths I think I understand the confusion on my original feedback that this should be saved to the form meta.

By default, WordPress' TinyMCE implementation does treat visual vs text editing as a user-specific preference. And it makes sense that we might want to preserve that. 

My concern is that toggling between these views can actually break custom markup as the visual editor will transform the text-based content into its own desired format.

This makes me want to break with the norm here and make the format (visual vs text) a field-specific setting rather than a user-specific preference.

Thoughts?

spivurno:
@malayladu What I'm saying is that local storage doesn't solve the problem we're trying to solve here. Local storage is inherently a user preference. Many different users will edit a given form and interact with these fields. If this is a user preference than the carefully crafted markup from User A could be unintentionally undone by User B just by loading the field settings for a Rich Text HTML field.

Some examples from the wild:

1. https://stackoverflow.com/questions/44000699/how-to-make-wordpress-editor-tinymce-preserve-html-formatting-when-switching-b
2. https://wordpress.stackexchange.com/questions/214588/tinymce-editor-is-breaking-my-beautiful-html
3. https://github.com/WordPress/classic-editor/issues/155

I don't think we'll need any fancy code to pull this off. Here's what I'm thinking.

1. Whenever the field is toggled to visual or text mode, we set a property on the field object via JS (no AJAX required).
2. When the form is saved, the current mode is saved as part of the field meta automatically (no extra code required).
3. When the field settings are loaded for an HTML field, the property is checked and the editor is loaded in the last saved mode based on the field property.

In this scenario, I don't think we need local storage at all.

Thoughts?

malayladu:
> @malayladu What I'm saying is that local storage doesn't solve the problem we're trying to solve here. Local storage is inherently a user preference. Many different users will edit a given form and interact with these fields. If this is a user preference than the carefully crafted markup from User A could be unintentionally undone by User B just by loading the field settings for a Rich Text HTML field.
> 
> Some examples from the wild:
> 
> 1. https://stackoverflow.com/questions/44000699/how-to-make-wordpress-editor-tinymce-preserve-html-formatting-when-switching-b
> 2. https://wordpress.stackexchange.com/questions/214588/tinymce-editor-is-breaking-my-beautiful-html
> 3. [HTML stripped when switching between Visual & Text WordPress/classic-editor#155](https://github.com/WordPress/classic-editor/issues/155)
> 
> I don't think we'll need any fancy code to pull this off. Here's what I'm thinking.
> 
> 1. Whenever the field is toggled to visual or text mode, we set a property on the field object via JS (no AJAX required).
> 2. When the form is saved, the current mode is saved as part of the field meta automatically (no extra code required).
> 3. When the field settings are loaded for an HTML field, the property is checked and the editor is loaded in the last saved mode based on the field property.
> 
> In this scenario, I don't think we need local storage at all.
> 
> Thoughts?

Ah, I see now. We want the editor mode setting to be based on the **field**, not the user. So once it's saved, any user who loads that field will see it in the `Visual` or `HTML` mode it was last saved in.

spivurno:
@malayladu Exactly! 😄

PR #1074, opened 2025-04-14
`gw-display-html-field-on-entry-detail.php`: Added a snippet to display HTML field content on Entry Details page.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2901917754/81943/

## Summary

Allow HTML field content to displayed on Entry Details page.

https://www.loom.com/share/fa4b49240b6447c7839392af69476bf2

### DIFF
--- a/gravity-forms/gw-display-html-field-on-entry-detail.php
+++ b/gravity-forms/gw-display-html-field-on-entry-detail.php
@@ -2,6 +2,8 @@
 /**
  * Gravity Wiz // Gravity Forms // Display HTML Field on Entry Details
  *
+ * Instruction Video: https://www.loom.com/share/fa4b49240b6447c7839392af69476bf2
+ *
  * Save and display HTML field content (including Live Merge Tags and shortcodes) in the entry detail view.
  * Useful for retaining dynamic HTML field output as it appeared when the entry was submitted.
  *
--- a/gravity-forms/gw-display-html-field-on-entry-detail.php
+++ b/gravity-forms/gw-display-html-field-on-entry-detail.php
@@ -1,0 +1,123 @@
+<?php
+/**
+ * Gravity Wiz // Gravity Forms // Display HTML Field on Entry Details
+ *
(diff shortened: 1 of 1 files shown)


spivurno:
@saifsultanc Just a note this should support the Entry List as well. 🙏

saifsultanc:
> @saifsultanc Just a note this should support the Entry List as well. 🙏

Yup. Done!

gravity-forms/gw-display-html-field-on-entry-detail.php
```diff
@@ -0,0 +1,128 @@
...
+				'field_id' => false,
+			)
+		);
+
+		add_action( 'init', array( $this, 'init' ) );
+	}
+
+	public function init() {
+		add_filter( 'gform_entry_post_save', array( $this, 'save_html_field_content' ), 10, 1 );
+		add_action( 'gform_entry_detail', array( $this, 'display_html_field_content_entry_detail' ), 10, 2 );
```
veryspry, line 35: What do you think about generating the HTML content on entry detail page load instead of persisting the info with the entry?

Two reasons:

* That would allow this to work on existing entries without needing to update the entry to trigger the HTML content generation.
* Updates to the snippet would then automatically apply rather than needing to go back through and manually updating / saving entries to get the updates reflected in the displayed data.

saifsultanc: @veryspry 🤔 From what I am able to recall, one reason for doing it this way to ensure dynamic loading of any updated HTML content.

veryspry: @saifsultanc wouldn't it work the opposite the way it is — like wouldn't the content only update after the entry is saved? Or is that okay as this is displaying HTML Field content that would only ever change if the field is updated?

I guess either way, I don't know if the additional `meta` is required here as the content is already available on the `$field` itself, right?

PR #1082, opened 2025-04-24
`gw-rich-text-html-fields.php`: Fixed an issue where Rich Text HTML snippet only worked with Gravity Perks active.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2913661025/82373

## Summary

The [Rich Text HTML Fields snippet](https://gravitywiz.com/snippet-library/gw-rich-text-html-fields/) doesn't work as expected unless the Gravity Perks base plugin is installed and active.

This fixes the issue.

### DIFF
--- a/gravity-forms/gw-rich-text-html-fields.php
+++ b/gravity-forms/gw-rich-text-html-fields.php
@@ -19,6 +19,12 @@ add_action( 'admin_init', function() {
 	}
 } );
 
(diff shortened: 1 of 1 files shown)


spivurno:
@saifsultanc A suggestion/explanation: https://www.loom.com/share/cbd4b3190fd04125956cba07917648cd

PR #1083, opened 2025-04-24
`gw-choice-counter.php`: Fixed an issue with Choice Counter not working for Image Choice field.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2917523082/82535

## Summary

The Choice Counter is not working with Image Choice Field. The count is 1 off; you select two, it shows 1. You select 3 it'll show 2.

This update addresses an off-by-one issue when counting selected options in Gravity Forms' Image Choice fields. The problem occurs because clicking on an image triggers a click on the associated label, which then toggles the underlying checkbox. However, the checkbox state wasn't immediately updated at the time of the event, so there was a slight delay before the DOM reflected the change.

Previously, the count was being calculated immediately on click, which sometimes captured the state before the checkbox had updated, leading to inaccurate totals.

**BEFORE:**
Screenshot 2025-04-24 at 2 12 40 PM

**AFTER:**
Screenshot 2025-04-24 at 2 12 31 PM

### DIFF
--- a/gravity-forms/gw-choice-counter.php
+++ b/gravity-forms/gw-choice-counter.php
@@ -90,9 +90,9 @@ class GW_Choice_Count {
 					// Event handler for all listeners to avoid DRY and to maintain a pointer reference to the function
 					// which we can use to explicity unbind event handlers
 					self.updateChoiceEventHandler = function() {
-						setTimeout( function() {
+						requestAnimationFrame( function() {
 							self.updateChoiceCount( self.formId, self.choiceFieldIds, self.countFieldId, self.values );
-						}, 10 ); // 10ms delay gives DOM time to reflect input change
+						}); // 10ms delay gives DOM time to reflect input change
 					};
 
 					self.init = function() {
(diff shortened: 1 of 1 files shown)


gravity-forms/gw-choice-counter.php
```diff
@@ -90,7 +90,9 @@ function output_script() {
 					// Event handler for all listeners to avoid DRY and to maintain a pointer reference to the function
 					// which we can use to explicity unbind event handlers
 					self.updateChoiceEventHandler = function() {
-						self.updateChoiceCount( self.formId, self.choiceFieldIds, self.countFieldId, self.values );
+						setTimeout( function() {
```
veryspry, line 93: What about the following?

I think it might be slightly more reliable as I believe it will ensure that the code runs after the DOM changes but before the next repaint

```suggestion
						requestAnimationFrame(function() {
							self.updateChoiceCount( self.formId, self.choiceFieldIds, self.countFieldId, self.values );
						});
```

PR #1085, opened 2025-04-25
`gpalf-count-only-non-blank-rows.js`: Added support to calculate only non-blank rows in a list field.

malayladu:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2916143012/82450

## Summary

By default, GP Auto List Field calculates the all rows (including blank rows) when we use `:count` modifier. This snippet is useful when you want to count only non-blank rows.

82450-Count-Non-Blank-Rows


saifsultanc (approved):
Tests look good. LGTM. S&M!

PR #1091, opened 2025-04-30
`gpml-acf-user-image-field.php`: Fixed an issue with User Image getting removed on form resubmission.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2921885802/82733

## Summary

File uploaded to an ACF Image field on the user profile, when user profile is updated with GFUR is removed.

https://www.loom.com/share/029fcc517e50428d81800577d1106661?sid=ac7c7d82-17ae-4161-bda6-f4c3478d5e50

### DIFF
--- a/gp-media-library/gpml-acf-user-image-field.php
+++ b/gp-media-library/gpml-acf-user-image-field.php
@@ -33,53 +33,50 @@ class GPML_ACF_User_Image_Field {
 
 	}
 
+	protected function get_attachment_id_by_filename( $filename ) {
+		global $wpdb;
+
+		return $wpdb->get_var( $wpdb->prepare(
+			"SELECT ID FROM $wpdb->posts WHERE post_type = 'attachment' AND guid LIKE %s LIMIT 1",
+			'%' . $wpdb->esc_like( $filename ) . '%'
+		));
+	}
+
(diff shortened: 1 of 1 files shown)


gp-media-library/gpml-acf-user-image-field.php
```diff
@@ -44,6 +44,31 @@ function update_user_image_field( $user_id, $feed, $entry ) {
...
+			$raw_json = $_POST['gform_uploaded_files'] ?? '';
+			if ( empty( $value ) && $raw_json ) {
+				$clean_json = stripslashes( $raw_json );
+
+				$uploaded_files_array = json_decode( $clean_json, true );
+
+				$field_id = $this->_args['field_id'];
+				$key      = 'input_' . $field_id;
+
+				if ( isset( $uploaded_files_array[ $key ][0]['uploaded_filename'] ) ) {
```
veryspry, line 57: @saifsultanc does this need to work with single file upload fields as well?

If so, you'll likely need to do a check on `$_FILES` if it's a single file upload field. You can check the field type like so: `$field->multipleFiles`

Here's an example of `$_FILES` with a file for a single file upload field.


image

veryspry (approved):

PR #1094, opened 2025-05-03
`gw-conditional-logic-operator-does-not-contain.php`: Fixed rule value for `does not contain` rule.

saifsultanc:
## Context

💬 Slack: https://gravitywiz.slack.com/archives/C04RQJ232PQ/p1746134330652079

## Summary

The does not contain rule renders select of dropdown, instead of the text box.

David's loom explaining the issue:
https://www.loom.com/share/fc3e8050e0284dc480247ad2e35112ff

This PR adds the override for `does_not_contain` operator as well. Loom Demo: https://www.loom.com/share/5eeed1dd533449eab06cef3088922def

I had initially thought of the approach of using 

```js
			gform.addFilter( 'gform_conditional_logic_values_input', function( str, objectType, ruleIndex, selectedFieldId, value ) {
				const operator = jQuery( '#field_rule_operator_' + ruleIndex ).val();
				if ( operator !== 'does_not_contain' ) {
					return str;
				}
				str = '<input type="text" placeholder="' + gf_vars['enterValue'] + '" data-js-rule-input="value" class="gfield_rule_select gfield_rule_input active" id="field_rule_value_' + ruleIndex + '" name="field_rule_value_' + ruleIndex + '" value="' + value + '">';
				return str;
			} );
```

However, the issue with this approach was that on first render the `field_rule_value_` was not defined yet, and we couldn't get `does not contain` to render the string instead of the dropdown of the values (for this example). It did work fine on frontend and any refresh on the values by adding/removing more rules. However, this solution was clearly lacking the correct behaviour. The issue with this approach:

https://www.loom.com/share/877e59bf2b9147e2aed59527754ec31e

### DIFF
--- a/gravity-forms/gw-conditional-logic-operator-does-not-contain.php
+++ b/gravity-forms/gw-conditional-logic-operator-does-not-contain.php
@@ -45,14 +45,11 @@ class GF_CLO_Does_Not_Contain {
 				operators.does_not_contain = 'does not contain';
 				return operators;
 			} );
-			gform.addFilter( 'gform_conditional_logic_values_input', function( str, objectType, ruleIndex, selectedFieldId, value ) {
-				const operator = jQuery( '#field_rule_operator_' + ruleIndex ).val();
-				if ( operator !== 'does_not_contain' ) {
-					return str;
-				}
-				str = '<input type="text" placeholder="Enter a value" data-js-rule-input="value" class="gfield_rule_select gfield_rule_input active" id="field_rule_value_' + ruleIndex + '" name="field_rule_value_' + ruleIndex + '" value="' + value + '">';
-				return str;
-			} );
+
+			// Override the default GF function to add our custom operator.
+			function ruleNeedsTextValue( rule ) {
+				return ['does_not_contain','contains', 'starts_with', 'ends_with', '<', '>' ].indexOf ( rule.operator ) !== -1;
+			}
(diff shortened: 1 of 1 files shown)


gravity-forms/gw-conditional-logic-operator-does-not-contain.php
```diff
@@ -45,6 +45,11 @@ public function output_admin_inline_script() {
 				operators.does_not_contain = 'does not contain';
 				return operators;
 			} );
+
+			// Override the default GF function to add our custom operator.
+			function ruleNeedsTextValue( rule ) {
+				return ['does_not_contain','contains', 'starts_with', 'ends_with', '<', '>' ].indexOf ( rule.operator ) !== -1;
+			}
```
spivurno, line 52:
```suggestion
			let origRuleNeedsTextValue = window.ruleNeedsTextValue;
			// Override the default GF function to add our custom operator.
			function ruleNeedsTextValue( rule ) {
			    let needsTextValue = origRuleNeedsTextValue( rule );
			    return needsTextValue || rule.operator.indexOf( 'does_not_contain' ) !== -1;
			}
```

Just to be good neighbors, I'd use this approach so that other people could override this function themselves and we if Gravity Forms adds future exceptions to the function, we don't interfere.

spivurno (approved):

PR #1099, opened 2025-05-08
`gpld-populate-new-minimum-date-into-linked-date-field.js`: Fixed an issue with default value of field notgetting used to apply minimum date.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2929104779/83029

## Summary

The snippet above doesn't work on initial load if the Date field A has a default date value.

Loom of the issue and update:

https://www.loom.com/share/763dfa9262a44ead841b3fdb2818fac4


veryspry (approved):
Just to confirm, does this still work if the source field does not have a default value?

If yes, this looks great to me

PR #1100, opened 2025-05-09
`gw-disable-submission-on-enter.js`: Fixed an issue with disable enter snippet not working with Conversational Forms.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2931919718/83124

## Summary

[Disable Submission on Enter](https://gravitywiz.com/snippet-library/gw-disable-submission-on-enter/) snippet and Gravity Forms Conversational Forms Add-on incompatibility. The snippet works as expected when Conversational forms is not active on the form, but once it is active on the form, the snippet doesn't work.

We have updated to
1. Use `addEventListener` with `keydown` to catch `Enter` across all input elements.
2. Check for `.gform_wrapper` as a parent to confirm the event is inside a Gravity Form field.
3. Add `stopImmediatePropagation()` to prevent interference from other listeners.

This works reliably with GF Conversational Forms Add-on, and continues to work with normal form preview/form front end pages.

**BEFORE** (Credits: Roxy):
https://www.loom.com/share/9c577fd7af1d47f090458e99e9118444

**AFTER**:
https://www.loom.com/share/050bd75bd63b475c9827a516083b4f84

### DIFF
--- a/gravity-forms/gw-disable-submission-on-enter.js
+++ b/gravity-forms/gw-disable-submission-on-enter.js
@@ -10,7 +10,21 @@
 document.addEventListener( 'keydown', function(e) {
 	var isGravityForm = e.target.closest( '.gform_wrapper' ) !== null;
 	var code = e.keyCode || e.which;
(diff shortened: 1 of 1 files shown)


veryspry (approved):
I like it 💯

PR #1101, opened 2025-05-09
`gpnf-hide-quantity-in-summary-field-values.php`: Added snippet to hide quantity from GPNF summary field values.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2929910662/83058

## Summary

To remove the quantity from the product data which is display on the Nested Forms Summary table.

The snippet is language independent!

**_Deutsch:_**
Screenshot 2025-05-10 at 12 00 08 AM

**_English:_**
Screenshot 2025-05-10 at 12 00 23 AM

### DIFF
--- a/gp-nested-forms/gpnf-hide-quantity-in-summary-field-values.php
+++ b/gp-nested-forms/gpnf-hide-quantity-in-summary-field-values.php
@@ -1,0 +1,17 @@
+<?php
+/**
+ * Gravity Perks // Nested Forms // Hide Quantity in Summary Field Values
+ * https://gravitywiz.com/documentation/gravity-forms-nested-forms/
+ *
+ * Product field display values are formatted as "Label, Quantity, Price". This snippet removes the quantity from the display value.
+ */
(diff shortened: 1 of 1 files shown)


gp-nested-forms/gpnf-hide-quantity-in-summary-field-values.php
```diff
@@ -0,0 +1,17 @@
+<?php
+/**
+ * Gravity Perks // Nested Forms // Hide Quantity in Summary Field Values
+ * https://gravitywiz.com/documentation/gravity-forms-nested-forms/
+ *
+ * Product field display values are formatted as "Label, Quantity, Price". This snippet removes the quantity from the display value.
```
veryspry, line 6: Are there any cases where the order could change? E.g. is that order configurable?

A more specific question is: is there a config somewhere we could grab and dynamically build the regex from that?

I wouldn't sink too much time into this, if any. More so just a question in case it's an easy thing to add

veryspry (approved):
@saifsultanc approved, with one question

PR #1107, opened 2025-05-22
`gpeb-multi-field-sorting.php`: Added snippet for multi field sorting on Entry Blocks.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2945420521/83832

## Summary

Add support for multi-level sorting in Entry Blocks, allowing users to define a primary and secondary sort field.

Demo:
https://www.loom.com/share/b9867d2735d44519bf563961e9b30bd2

### DIFF
--- a/gp-entry-blocks/gpeb-multi-field-sorting.php
+++ b/gp-entry-blocks/gpeb-multi-field-sorting.php
@@ -1,0 +1,41 @@
+<?php
+/**
+ * Gravity Perks // Entry Blocks // Multi Field Sorting
+ * https://gravitywiz.com/documentation/gravity-forms-entry-blocks/
+ *
+ * Use confirmation message from Form Settings when editing an entry in GPEB.
+ * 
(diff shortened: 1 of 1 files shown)


gp-entry-blocks/gpeb-multi-field-sorting.php
```diff
@@ -0,0 +1,41 @@
...
+ * https://gravitywiz.com/documentation/gravity-forms-entry-blocks/
+ *
+ * Use confirmation message from Form Settings when editing an entry in GPEB.
+ * 
+ * Instruction Video: https://www.loom.com/share/b9867d2735d44519bf563961e9b30bd2
+ */
+add_filter( 'gpeb_queryer_entries', function( $entries, $gf_queryer ) {
+
+	// Update the form ID to match your form.
+	if ( $gf_queryer->form_id != 933 ) {
```
veryspry, line 13: Can you store the form id in a variable? (I think that's how we usually do it)

PR #1123, opened 2025-06-20
`gw-product-quantity-conditional.php`: Added snippet to use quantity value of Product Field as a conditional logic field.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2951687684/84205

## Summary

Targeting Single Product field's quantity input to be applicable in Gravity Forms' regular conditional logic.

Snippet Demo:
https://www.loom.com/share/38547a28853f4706b10cfdb42eaa45ca

### DIFF
--- a/gravity-forms/gw-product-quantity-conditional.php
+++ b/gravity-forms/gw-product-quantity-conditional.php
@@ -1,0 +1,105 @@
+<?php
+/**
+ * Gravity Wiz // Gravity Forms // Product Quantity Conditional
+ *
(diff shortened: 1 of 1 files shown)


spivurno:
@saifsultanc Does this work on submission as well? I was expecting to see some PHP validation but maybe it's not required?

veryspry (approved):
Nice!

Quick question: can these methods be removed? (The don't seem to be in use)

- `isCustomQtyField()`
- `getCustomQtyFieldId()`
- `isQtyField()`

PR #1132, opened 2025-07-17
`gw-rich-text-html-fields.php`: Added improvements for Rich Text HTML Fields Snippet.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2987432785/86002

## Summary

Primarily inspired from https://github.com/gravitywiz/snippet-library/pull/1071, with some improvements. Takes into account the recent updates to the snippet here https://github.com/gravitywiz/snippet-library/commit/55c24385053d8d546c36abd6e4b285b0dc55f4c8

https://www.loom.com/share/fbb020ebea884e76a1ed2fe0eae1e9fb

### DIFF
--- a/gravity-forms/gw-rich-text-html-fields.php
+++ b/gravity-forms/gw-rich-text-html-fields.php
@@ -60,6 +60,20 @@ add_action( 'gform_field_standard_settings', function( $position ) {
 	</style>
 
 	<script>
+		// Immediately disable HTML click.
+		jQuery( '.gf-html-container' )
+			.css( 'pointer-events', 'none' )
+			.css( 'opacity', '0.6' );
+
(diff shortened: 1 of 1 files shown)


spivurno:
@saifsultanc A few issues:

1. Code is being wrapped in tags automatically on save:
    https://www.loom.com/share/327f74651a0d4f2da0f3e45d74eda7e4
    
2. Line breaks are being removed in the code editor. 
    https://www.loom.com/share/d46948fae6f04518bf45a81442ebe5c8
    
    I believe this is an extension of the issue I reported originally where the code editor does not save values with line breaks. See the original Loom for details on that.
    
3. The issue where the height of the code editor increases each time the field settings were saved is still present. See the original Loom for details.

PR #1135, opened 2025-07-21
`gpnf-sort-nested-form-entries.php`: Fixed date field sorting for European formats

SebastianWiz:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/2995555966/86254?viewId=8172236

## Summary

**Issue:** Date field sorting failed when using European formats like `dd.mm.yyyy`

**Fix:** Use the standardized `value` property for date fields instead of the formatted `label` property.


saifsultanc (approved):
LGTM!

PR #1142, opened 2025-08-07
`gpuid-multiple-ids.php`: Added option to use multiple fields for count

codepuncher:
## Context

https://gravitywiz.com/gravity-forms-build-raffle-or-lottery/

## Summary

This is to extend the raffle/lottery ticket example on here: https://gravitywiz.com/gravity-forms-build-raffle-or-lottery/

While the guide supports "packages" of pre-defined counts for choice selections, it does not allow user-defined counts for multiple ticket types.

This PR allows you to specify multiple field IDs that will all be counted and totalled to the final count allowing you to sell multiple ticket types with as many as a user would like to purchase.


veryspry (approved):
Thanks for the submission!

PR #1144, opened 2025-08-11
`gw-custom-modifier-file-upload.php`: Added snippet for custom `filename` modifier for file upload fields.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/3025552109/87377

https://github.com/gravitywiz/gp-media-library/pull/35#issuecomment-3156459815

## Summary

This snippet is an example of how to add custom modifiers to a Gravity Forms file upload field

Also dependent on GPML https://github.com/gravitywiz/gp-media-library/pull/35


spivurno:
@saifsultanc I'd recommend making this snippet title specific to the `:filename` modifier. If we need to add more modifiers in the future we can either update the snippet then or create new ones.

PR #1167, opened 2025-09-17
`gpapf-show-all-countries.php`: Added new snippet.

matty0501:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/3078299767/89409?viewId=7627047

## Summary

This snippet will set the allowed countries in GPAPF to `All` for a specific form/field. Useful for if a user has locked down the countries in the plugin settings, but wants to allow all countries for a specific form.


saifsultanc (approved):
Nice!

PR #1172, opened 2025-09-26
`gpb-daily-service-booking-limit.php`: Added new snippet

SebastianWiz:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/3084918923/89679?viewId=8172236

## Summary

This PR adds a snippet that lets you set a daily booking limit for one or multiple services. The snippet hooks into Gravity Forms validation to block submissions once the combined quantity for that day reaches the limit and shows the “fully booked” validation message on the booking-time field. It also hooks `gpb_before_booking_created` so the same check runs during booking creation, preventing overbookings if multiple requests land at the same time.

Loom demo: https://www.loom.com/share/6cae41c053a647e48deec68868e23ef4

### DIFF
--- a/gp-bookings/gpb-daily-service-booking-limit.php
+++ b/gp-bookings/gpb-daily-service-booking-limit.php
@@ -1,0 +1,213 @@
+<?php
+/**
+ * Gravity Perks // GP Bookings // Daily Service Booking Limit
+ * https://gravitywiz.com/documentation/gravity-forms-bookings/
+ *
+ * Enforce a daily capacity for one or more GP Bookings services. When the selected services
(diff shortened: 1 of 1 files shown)


SebastianWiz:
@spivurno Thank you! I've been exploring this and I think we could hook into the REST API endpoint that provides calendar availability to mark days as unavailable when they hit the limit. That way we could prevent the days from being selectable in the first place rather than showing a validation error after submission. That would definitely be better UX for sure. 

Should hopefully have an update to the snippet soon!

saifsultanc (approved):

PR #1236, opened 2026-02-18
`gpeb-edit-entry-by-unique-id.php`: Added new snippet to allow unauthenticated entry editing via GP Unique ID as access token.

SebastianWiz:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/3231888116/98034?viewId=3808239

## Summary

This PR adds a new snippet that allows non-logged-in users to edit entries using a GP Unique ID as an access token. Instead of using the numeric entry ID in the URL, the unique ID identifies the entry and grants edit access, so only someone with the correct unique ID link can edit that entry.

### DIFF
--- a/gp-entry-blocks/gpeb-edit-entry-by-unique-id.php
+++ b/gp-entry-blocks/gpeb-edit-entry-by-unique-id.php
@@ -1,0 +1,201 @@
+<?php
+/**
+ * Gravity Perks // Entry Blocks // Edit Entry by GP Unique ID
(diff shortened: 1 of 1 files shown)


saifsultanc (changes requested):
One smoll Comment and can we add a loom instructions to the snippet header?

saifsultanc:
@SebastianWiz Bump on this one!

Can you please get this cleaned up, and re-tested on GF3.x, and update (if necessary). This is a good one to add to the SL! 🥇

PR #1263, opened 2026-04-07
`gpp-pass-trigger-data-attributes-to-popup-form.php`: Added new snippet for passing trigger data attributes to popup form.

SebastianWiz:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/3281164510/100352?viewId=3808239

## Summary

**Note: this snippet is in draft until https://github.com/gravitywiz/gp-popups/pull/17 is merged.**

This PR adds a new snippet to pass data-* attributes from popup trigger elements into the popup form via dynamic population. 

Useful when you have multiple triggers pointing to the same popup but need to pass different values depending on which trigger was clicked (e.g. a product ID, event ID, etc).

The snippet captures data-* attributes from the clicked trigger during the capture phase, then injects them as query params on the popup iframe URL using the `gpp_popup_config` filter.

### DIFF
--- a/gp-popups/gpp-pass-trigger-data-attributes-to-popup-form.php
+++ b/gp-popups/gpp-pass-trigger-data-attributes-to-popup-form.php
@@ -1,0 +1,127 @@
+<?php
+/**
+ * Gravity Perks // Popups // Pass Trigger Data Attributes to Popup Form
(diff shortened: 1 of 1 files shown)


SebastianWiz:
Snippet now ready for review since https://github.com/gravitywiz/gp-popups/pull/17 has been merged 👍

PR #1267, opened 2026-04-15
`gpld-exclude-blocked-dates-from-modifiers.php`: Fixed an issue where the minimum date overshot by one day when the reference date fell on a blocked day.

SebastianWiz:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/3285849749/100533?viewId=3808239

## Summary

**Issue**: When the reference date fell on a blocked day, the minimum date would overshoot by one working day. This happened because the blocked-date counting loop includes the base date, but in this case the base isn't part of the modifier range, so a blocked base gets counted when it shouldn't be.

**Fix**: Subtract the extra count when the base date is blocked. Applied to both PHP and JS.


saifsultanc (approved):
LGTM!

PR #1269, opened 2026-04-15
`gpld-us-federal-holidays.php`: Added new snippet for adding US federal holidays as exceptions.

SebastianWiz:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/3289096722/100633?viewId=3808239

## Summary

This PR standardizes Clifford's US Federal Holidays [snippet](https://gist.github.com/cliffordp/38892bbfe30702d4623c9c13e17a1b09) for the snippet library. Adds a credit link in the header, aligns class/method naming and formatting with our conventions, and restructures the constructor to use `wp_parse_args` with an `init` hook for filter registration.

### DIFF
--- a/gp-limit-dates/gpld-us-federal-holidays.php
+++ b/gp-limit-dates/gpld-us-federal-holidays.php
@@ -6,20 +6,20 @@
  * Add US federal holidays as exceptions to your GP Limit Dates fields. Supports fixed-date holidays (e.g.
  * Independence Day) and floating holidays (e.g. Thanksgiving, Memorial Day). Fixed holidays that fall on a
  * Saturday or Sunday are automatically shifted to the observed weekday (Saturday to Friday, Sunday to Monday).
- * 
+ *
  * Credit: Clifford (https://github.com/cliffordp)
  *
  * Instructions:
- * 
+ *
  * 1. Install this snippet by following the steps here:
  *    https://gravitywiz.com/documentation/how-do-i-install-a-snippet/
- * 
+ *
  * 2. Update the "targets" configuration at the bottom of this file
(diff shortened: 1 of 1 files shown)


gp-limit-dates/gpld-us-federal-holidays.php
```diff
@@ -0,0 +1,116 @@
...
+		return $holidays;
+	}
+
+}
+
+# Configuration
+
+new GPLD_US_Federal_Holidays( array(
+	'targets'           => array(
+		'123_4', // Form ID 123, Field ID 4
```
spivurno:
```suggestion
		123 => 4, // Form ID 123, Field ID 4
```

Typically, we'd do something like this. Any objections?

spivurno (changes requested):

gp-limit-dates/gpld-us-federal-holidays.php
SebastianWiz: @spivurno Ah yeah, you're right! Updated the snippet to use our more common pattern:

```
new GPLD_US_Federal_Holidays( array(
    'form_id'           => 123,
    'field_ids'         => array( 4, 5 ),
    'years_to_generate' => 20,
) );
```

spivurno: I dig it. Two last minute improvements:

1. If no `field_ids` are passed, apply to all fields for the specified form.
2. If no `form_id` is passed, apply to all fields on all forms.

SebastianWiz: @spivurno Good call! Updated

spivurno (approved):

PR #1272, opened 2026-04-20
`gpadvs-auto-select-on-match.js`: Fixed issue where auto select did not trigger when choices loaded after the user finished typing.

SebastianWiz:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/3294610193/100872?viewId=3808239

## Summary

Issue: The Auto Select on Match snippet only checked for a single match in `onType`, which fires before GPPA's AJAX request completes. If the user typed faster than choices could populate, the check ran against an empty or stale option set and never auto-selected.

Fix: wrap `onLoad` (which fires after Tom Select's lazy-load AJAX callback completes) to re-run the same match check against the now-populated options, using the user's last typed query.

Also fixed snippet's inconsistent formatting.


saifsultanc (approved):
Works Great! LGTM!

PR #1277, opened 2026-04-24
`gpppw-round-up-pricing.php`: Added snippet to round up pricing values.

saifsultanc:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/3300298065/101092?viewId=3808239

## Summary

Snippet to round up pricing values.

### DIFF
--- a/gp-pay-per-word/gpppw-round-up-pricing.php
+++ b/gp-pay-per-word/gpppw-round-up-pricing.php
@@ -1,0 +1,210 @@
+<?php
+/**
+ * Gravity Wiz // Pay Per Word // Round Up Price
+ * https://gravitywiz.com/documentation/gravity-forms-pay-per-word/
+ *
+ * Rounds up the Pay Per Word calculated price to the nearest whole dollar amount.
+ *
+ * Plugin Name:  GP Pay Per Word - Round Up Price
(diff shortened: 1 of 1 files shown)


spivurno:
@saifsultanc Looking good, man! One consideration for you… curious how well you think this might fit into our [GF Rounding](https://gravitywiz.com/gravity-forms-rounding/) plugin? Seems like a strong place for this code to live. If you think it’s too messy, I’m down to roll with this. Just looking for opportunities to make our solutions more robust — where it makes sense.

saifsultanc:
> @saifsultanc Looking good, man! One consideration for you… curious how well you think this might fit into our [GF Rounding](https://gravitywiz.com/gravity-forms-rounding/) plugin? Seems like a strong place for this code to live. If you think it’s too messy, I’m down to roll with this. Just looking for opportunities to make our solutions more robust — where it makes sense.

@spivurno Indeed! But it will need some updates to GF Rounding as per : https://github.com/gravitywiz/gf-rounding/blob/master/gf-rounding.php#L150-L159

If you suggest I will drop this, and spend some more time investigating incorporating the logic to GF Rounding instead!

spivurno:
Haha, I found myself waffling back and forth a bunch on this, which makes me think that we should probably just move forward with this snippet. Ignore me! 😆

PR #1294, opened 2026-05-27
`gw-edit-product-and-payment-details.php`: Added support for removing disabled-quantity products when editing entries.

SebastianWiz:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/3333283965/102854?viewId=3808239

## Summary

This PR adds support for removing Single Product fields with quantity disabled when editing a Gravity Forms entry.

Previously, admins could remove products from an entry by setting quantity to `0` only when the product field had an editable quantity input, either from quantity being enabled on the product field or from a standalone Quantity field. This is useful if an admin, for example, has issued a refund via Stripe and wants to update the entry to match. However, Single Product fields with quantity disabled did not expose a quantity input and therefor these products could not be removed from an entry.

Now the snippet detects disabled-quantity Single Product fields with no mapped Quantity field and temporarily sets `disableQuantity` to `false` while Gravity Forms renders the input, so the normal quantity field appears (the quantity input is limited to 0 or 1).

### DIFF
--- a/gravity-forms/gw-edit-product-and-payment-details.php
+++ b/gravity-forms/gw-edit-product-and-payment-details.php
@@ -4,7 +4,7 @@
  *
  * Edit products (and payment details) via the Gravity Forms Edit Entry view.
  *
- * @version   1.3
+ * @version   1.4
  * @author    David Smith <david@gravitywiz.com>
  * @license   GPL-2.0+
  * @link      http://gravitywiz.com/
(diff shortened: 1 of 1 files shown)


spivurno:
@SebastianWiz Just stopped by to make sure that we were accounting for the scenario where the quantity was disabled on the field, but a quantity field existed—and you already did. 😊

SebastianWiz:
@spivurno Great minds think alike 🤝! 😆

saifsultanc (approved):
Perfecto!

PR #1300, opened 2026-06-02
`gpcp-show-original-price.js`: Fixed display issues and added support for GPPA dynamic pricing.

SebastianWiz:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/3338825288/103089?viewId=3808239

## Summary

Issue: the crossed-out price rendered  above the field instead of inline, and it showed even when no discount applied (the wrapper's text includes the label/quantity, so the comparison never matched). It also only ever added the crossed-out element, so it didn't stay in sync as prices changed. Separately, it read the original  price from `getBasePrice()`, which only holds the static prices captured at page render, so when GP Populate Anything set the price live, the crossed-out value was stale.

Fix: now targets the price element directly (`span#input_*`, or the readonly `input#ginput_base_price_*` when Quantity is disabled), fixing the positioning and comparison, and refreshes the crossed-out text on every update. For live pricing, it captures each product's base price from GPPA's `gppa_updated_batch_fields` event and falls back to the static price otherwise.

### DIFF
--- a/gp-conditional-pricing/gpcp-show-original-price.js
+++ b/gp-conditional-pricing/gpcp-show-original-price.js
@@ -1,34 +1,77 @@
 /**
  * Gravity Perks // GP Conditional Pricing // Show the original price crossed out next to the adjusted price
- * http://gravitywiz.com/documentation/gravity-forms-conditional-pricing/
+ * https://gravitywiz.com/documentation/gravity-forms-conditional-pricing/
  *
  * Instructions:
  *     1. Install our free Custom Javascript for Gravity Forms plugin.
  *        Download the plugin here: https://gravitywiz.com/gravity-forms-code-chest/
  *     2. Copy and paste the snippet into the editor of the Custom Javascript for Gravity Forms plugin.
  */
+var gwcpBasePrices = {};
+
+$( document ).off( 'gppa_updated_batch_fields.gwcp' ).on( 'gppa_updated_batch_fields.gwcp', function( event, formId, updatedFieldIds ) {
+
+	var gwcp = GWConditionalPricing.getFormElement( formId ).data( 'gwcp' );
+	if ( ! gwcp || ! gwcp._pricingLogic || ! updatedFieldIds || ! updatedFieldIds.length ) {
+		return;
+	}
+
+	var currency = typeof gform !== 'undefined' && typeof gform.Currency === 'function'
+		? new gform.Currency( gf_global['gf_currency_config'] )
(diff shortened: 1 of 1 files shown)


gp-conditional-pricing/gpcp-show-original-price.js
```diff
@@ -1,34 +1,77 @@
...
-	else if ( ! $basePrice.length ) {
-		$basePrice = '<span id="base_price_{0}_{1}" style="text-decoration:line-through;margin-right:0.3rem">{2}</span>'.gformFormat( GWConditionalPricing._formId, productId, basePrice );
-		$input.before( $basePrice );
+	else {
+		if ( ! $basePrice.length ) {
+			$basePrice = '<span id="base_price_{0}_{1}" style="text-decoration:line-through;margin-right:0.3rem"></span>'.gformFormat( GWConditionalPricing._formId, productId );
+			$input.before( $basePrice );
+		}
+
+		$basePrice.text( basePrice );
```
saifsultanc: I'd a check to prevent some console error that were popping up

```suggestion
		if ( $basePrice && typeof $basePrice.text === 'function' ) {
			$basePrice.text( basePrice );
		}
```

gp-conditional-pricing/gpcp-show-original-price.js
```diff
@@ -1,34 +1,77 @@
...
 		return;
 	}
 
+	var displayedPrice = isInput ? $input.val() : $input.text();
+
 	var $basePrice = $( '#base_price_{0}_{1}'.gformFormat( GWConditionalPricing._formId, productId ) );
 
 	// Remove base price if there is no discounted price.
-	if ( $input.text() === basePrice ) {
+	if ( displayedPrice === basePrice ) {
```
saifsultanc, line 64: This doesnt seem to be firing when we switch off the conditional pricing. The fallback to base price should look elegant, if possible.

This still shows the crossed out price, but at this instant the price is actually 30. 

Image

Think this isnt an issue for customer scenario, but lets see if we can fix it?

saifsultanc (changes requested):
Nice work! A couple of suggestions, so I can recheck 👀

gp-conditional-pricing/gpcp-show-original-price.js
SebastianWiz: @saifsultanc This was a bug in the core plugin, fix here: https://github.com/gravitywiz/gwconditionalpricing/pull/90

saifsultanc (approved):
LGTM!

PR #1303, opened 2026-06-05
`gpb-minimum-qty-by-remaining-capacity.php`: Added new snippet for setting a minimum qty by remaining capacity.

SebastianWiz:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/3344406510/103310?viewId=3808239

## Summary

Added a new snippet for specifying a minimum required booking quantity, based on the remaining capacity.

When a slot's remaining capacity is at or above a configurable threshold, the customer must book at least a minimum quantity. Once enough spots are taken that  remaining capacity drops below the threshold, the minimum no longer applies.


saifsultanc (approved):
Nice! LGTM

PR #1305, opened 2026-06-09
`gw-update-posts.php`: Fixed issue where file upload URLs were saved with surrounding brackets and quotes on Gravity Forms 2.10.

SebastianWiz:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/3345214180/103343?viewId=3808239

## Summary

Issue: Mapping a file upload field to a post custom field stored a JSON-encoded array (e.g. `["https://example.com/file.jpg"]`) instead of the plain URL. As of Gravity Forms 2.10, file upload fields store their value as JSON even for single uploads, and the snippet was copying that raw entry value straight into the custom field.

Fix: The snippet now resolves file upload values through the field's `get_value_export()` method, returning the plain URL (or a delimited list for multi-file uploads). This works across Gravity Forms versions, handling both legacy plain-URL and 2.10 JSON storage.

### DIFF
--- a/gravity-forms/gw-update-posts.php
+++ b/gravity-forms/gw-update-posts.php
@@ -186,7 +186,7 @@ class GW_Update_Posts {
 					}
 
 					// If the taxonomy is not hierarchical, we need to get the term names from the term ids.
-					if( ! $this->gw_is_taxonomy_hierarchical( $taxonomy ) ) {
+					if ( ! $this->gw_is_taxonomy_hierarchical( $taxonomy ) ) {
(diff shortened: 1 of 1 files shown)


gravity-forms/gw-update-posts.php
saifsultanc: @SebastianWiz Though not directly part of this PR / update, but it would be great if you can fix this linting issues. Rest look fine!

saifsultanc (changes requested):

PR #1316, opened 2026-06-15
`gpb-conditional-conferencing-by-field-value.php`: Added new snippet for conditional conferencing by field value.

SebastianWiz:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/3349810804/103527?viewId=8172236

## Summary

This PR adds a new GP Bookings snippet that makes Google Calendar conferencing conditional on an entry field value, instead of  attaching it to every booking.

When the GC Google Calendar integration has Conferencing set to Google Meet, a link is added to every pushed event. This snippet hooks the `gpb_google_calendar_event_data` filter and removes the conferencing data unless a configured field matches an allowed value, scoped to a specific form.

### DIFF
--- a/gp-bookings/gpb-conditional-conferencing-by-field-value.php
+++ b/gp-bookings/gpb-conditional-conferencing-by-field-value.php
@@ -1,0 +1,41 @@
+<?php
+/**
+ * Gravity Perks // Bookings // Conditional Conferencing by Field Value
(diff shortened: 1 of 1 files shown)


saifsultanc (approved):
Cleaned up and cleared DangerJS issues. LGTM!

PR #1319, opened 2026-06-27
`gpb-delay-booking-completed-notifications.php`: Added new snippet for delaying Booking Completed notifications.

SebastianWiz:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/3355764968/103766?viewId=8172236

## Summary

This PR adds a snippet that delays one or more **Booking Completed** notifications so they send a configurable amount of time after a booking ends, instead of immediately when the booking completes.

The snippet uses a `gform_disable_notification` filter to suppress the listed notifications when GP Bookings tries to send them on completion, and at that same point schedules a one-off Action Scheduler job (`gpb_send_delayed_completed_notification`) for `now + delay`. When the job runs it re-checks the booking is still a confirmed service booking, guards against duplicates with an entry-meta flag, then sends via `GFCommon::send_notifications()`.

### DIFF
--- a/gp-bookings/gpb-delay-booking-completed-notifications.php
+++ b/gp-bookings/gpb-delay-booking-completed-notifications.php
@@ -1,0 +1,137 @@
+<?php
+/**
+ * Gravity Perks // Bookings // Delay Booking Completed Notifications
+ * https://gravitywiz.com/documentation/gravity-forms-bookings/
+ *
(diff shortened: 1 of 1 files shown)


saifsultanc (changes requested):
Request to fix the PHPCS issues on this. Rest LGTM!

PR #1321, opened 2026-07-09
`gpb-form-specific-event-title-description.php`: Added new snippet for form-specific Event Title & Description.

SebastianWiz:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/3378281127/104640?viewId=3808239

## Summary

Event settings (event title/description with merge tag support) are configured on the service, so every booking form that books that service gets the same template. When one service is shared across multiple forms whose fields differ, a single template can't work, the same field ID (`{:3}`) means different things on each form.

This PR adds a new snippet that overrides the Google Calendar event title and/or description on a per service + form basis, using the existing `gpb_google_calendar_event_data` filter.

### DIFF
--- a/gp-bookings/gpb-form-specific-event-title-description.php
+++ b/gp-bookings/gpb-form-specific-event-title-description.php
@@ -1,0 +1,66 @@
+<?php
+/**
+ * Gravity Perks // Bookings // Form-Specific Event Title & Description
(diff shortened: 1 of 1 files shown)


saifsultanc (approved):
This just needed some DangerJs cleanup. Rest its smooth!

LGTM!

PR #1327, opened 2026-07-21
`gw-display-navigation-buttons-inline.css`: Added new snippet for displaying page navigation buttons inline.

SebastianWiz:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/3209016871/96688?viewId=8172236

## Summary

Thiis PR adds a CSS snippet that lets users display a form page's Previous/Next navigation inline with the last row of fields (instead of stacked below).

image


saifsultanc (approved):
Fixed DangerJS issues. Its looks great! 🙌

PR #1335, opened 2026-08-08
`gcoai-limit-generations-per-visitor.php`: Added new snippet for limiting generations per visitor.

SebastianWiz:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/3407682325/105817?viewId=3808239

## Summary

This PR adds a new snippet for GCOAI that caps how many times a visitor can trigger an OpenAI request from an OpenAI Stream or OpenAI Image field. The check runs on the AJAX endpoints at priority 1, ahead of GCOAI's own handlers, so it cannot be bypassed by scripted or repeated requests.

### DIFF
--- a/gc-openai/gcoai-limit-generations-per-visitor.php
+++ b/gc-openai/gcoai-limit-generations-per-visitor.php
@@ -73,14 +73,19 @@ class GCOAI_Limit_Generations {
 	}
 
 	private function is_applicable( $form_id, $field_id ) {
-		foreach ( array( 'form_id' => $form_id, 'field_id' => $field_id ) as $arg => $value ) {
+		$checks = array(
+			'form_id'  => $form_id,
+			'field_id' => $field_id,
+		);
+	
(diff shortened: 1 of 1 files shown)


gc-openai/gcoai-limit-generations-per-visitor.php
```diff
@@ -0,0 +1,151 @@
...
+
+		if ( $count >= (int) $this->args['limit'] ) {
+			$this->block( $this->args['message'] );
+		}
+
+		set_transient( $key, $count + 1, (int) $this->args['window'] );
+	}
+
+	private function is_applicable( $form_id, $field_id ) {
+		foreach ( array( 'form_id' => $form_id, 'field_id' => $field_id ) as $arg => $value ) {
```
saifsultanc, line 76: @SebastianWiz This is looking real good! Just a few things to polish, PHPCS wise.

saifsultanc (approved):

PR #1351, opened 2026-09-18
`gspc-bypass-checkout-zero-value-products.php`: Added new snippet for bypassing checkout on zero-value products.

SebastianWiz:
## Context

⛑️ Ticket(s): https://secure.helpscout.net/conversation/3414114106/106101?viewId=8172236

## Summary

When a product with a GSPC form is added to the cart and the cart needs neither payment nor shipping, the snippet creates the WooCommerce order right there on the `woocommerce_add_to_cart_redirect` filter and sends the customer to a confirmation page instead of the cart/checkout. It goes through `WC_Checkout::create_order()` so GSPC's own hooks still link the entries to the order items, and `payment_complete()` triggers the usual entry payment status update and delayed feeds.

Two settings: `form_id` limits the bypass to one form (or an array of forms), and `confirmation_page_id` swaps WooCommerce's order-received page for a page of the customer's choosing. Both default to off.


saifsultanc (approved):
Noice!

